Issue #3, opened 2016-08-10
Support custom request id


marcbachmann:
I'm already using a request id on req.id (gets set by external load balancer).
This module overwrites that one with an incrementing id, which results in duplicated request ids if I have multiple processes running.

```js
if (!req.id) {
  req.id = ++nextId
  nextId = ...
}
```
Do you think that is good enough?

davidmarkclements:
agreed this is needed - will accept PR :) - currently on low bandwidth - out of curiosity what load balancer does this?

marcbachmann:
I'll create a pr.

> out of curiosity what load balancer does this?

Heroku uses the http header `X-Request-ID`
https://devcenter.heroku.com/articles/http-request-id.

haproxy also supports a unique header
http://cbonte.github.io/haproxy-dconv/configuration-1.5.html#4-unique-id-header

PR #4, opened 2016-08-11
Reuse req.id if present

marcbachmann:
Requested in #3

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -18,7 +18,7 @@ function pinoLogger (opts, stream) {
 
   loggingMiddleware.logger = logger
 
-  var nextId = 0
+  var nextReqId = 0
 
   stream = logger.stream
 
@@ -48,8 +48,10 @@ function pinoLogger (opts, stream) {
 
   function loggingMiddleware (req, res, next) {
     var startTime = Date.now()
-    req.id = ++nextId
-    nextId = nextId % maxInt
+    if (req.id === undefined) {
+      req.id = ++nextReqId
+      nextReqId = nextReqId % maxInt
+    }
 
     var child = logger.child({req: req})
 
--- a/test.js
+++ b/test.js
@@ -83,6 +83,28 @@ test('allocate a unique id to every request', function (t) {
   })
 })
 
+test('reuses existing req.id if present', function (t) {
+  t.plan(2)
+
+  var dest = split(JSON.parse)
+  var logger = pinoLogger(dest)
+  var someId = 'id-to-reuse-12345'
+
+  function loggerWithExistingReqId (req, res) {
+    req.id = someId
+    logger(req, res)
+  }
+
+  setup(t, loggerWithExistingReqId, function (err, server) {
+    t.error(err)
+    doGet(server)
+  })
+
+  dest.on('data', function (line) {
+    t.equal(line.req.id, someId)
+  })
+})
+
 test('responseTime', function (t) {
   var dest = split(JSON.parse)
   var logger = pinoLogger(dest)


davidmarkclements:
1. does the extra closure var affect benchmarks?
2. this will still overwrite the pre-existing id - maybe we should include that somehow (`originalId` or `_id`  or something)

cc @mcollina

mcollina:
@davidmarkclements:

> 1. does the extra closure var affect benchmarks?
2. this will still overwrite the pre-existing id - maybe we should include that somehow (originalId or _id or something)

I think you should go through the PR again, there is no extra `var` and no overwrite if there is an `id`.

LGTM

marcbachmann:
Yes, that `if` statement shouldn't have an impact.

There's still the question whether we'd like to save the previous `req.id` value somewhere.
But then you can ask the same about the variables `res.startTime` and `req.log`.
So in general I wouldn't do it. We just have to document it properly.

IMO it would be also fine if it gets removed completely.
There's no reason to set it if it isn't used in here. ;)

mcollina:
@marcbachmann I'm lost :). Where is it overwritten?

davidmarkclements:
sorry not overwritten or extra var, I misread in a rush

@marcbachmann re "shouldn't have an impact" what do benchmarks show?

marcbachmann:
> @marcbachmann I'm lost :). Where is it overwritten?

I meant that one. https://github.com/davidmarkclements/pino-http/pull/4/files#diff-7f28009834c7b72b5b0080bc0487ccddL51

Doesn't it really belong into this module? For convenience, yes, that's nice. But in a few cases people might have a custom function. And in that case it's faster to just hook that up on their own.


@davidmarkclements 
> "shouldn't have an impact" what do benchmarks show?

You can't really test such detailed operations using the current benchmark setup.
But here are the node http server benchmarks ;)
I've used `./benchmark/pino-server.js`

reuse-request-id branch
```bash
$ autocannon http://localhost:3000 -c 50
Running 10s test @ http://localhost:3000
50 connections

Stat         Avg       Stdev   Max
Latency (ms) 6.73      6.73    162
Req/Sec      6902      898.91  7823
Bytes/Sec    766.77 kB 96.6 kB 884.74 kB

69k requests in 10s, 7.66 MB read

$ autocannon http://localhost:3000 -c 50
Running 10s test @ http://localhost:3000
50 connections

Stat         Avg       Stdev     Max
Latency (ms) 6.13      9.12      285
Req/Sec      7521.55   1528.49   9079
Bytes/Sec    836.33 kB 169.71 kB 1.02 MB

83k requests in 11s, 9.18 MB read

$ autocannon http://localhost:3000 -c 50
Running 10s test @ http://localhost:3000
50 connections

Stat         Avg       Stdev     Max
Latency (ms) 6.37      6.8       265
Req/Sec      7286.9    1359.8    8743
Bytes/Sec    805.27 kB 151.23 kB 983.04 kB

73k requests in 10s, 8.09 MB read

$ autocannon http://localhost:3000 -c 50
Running 10s test @ http://localhost:3000
50 connections

Stat         Avg       Stdev    Max
Latency (ms) 6.07      1.47     42
Req/Sec      7670.8    180.07   7963
Bytes/Sec    851.97 kB 16.38 kB 884.74 kB

77k requests in 10s, 8.51 MB read
```

master branch
```bash
$ autocannon http://localhost:3000 -c 50
Running 10s test @ http://localhost:3000
50 connections

Stat         Avg       Stdev    Max
Latency (ms) 7.04      1.83     53
Req/Sec      6560.91   841.4    7803
Bytes/Sec    725.36 kB 94.01 kB 884.74 kB

72k requests in 11s, 8.01 MB read

$ autocannon http://localhost:3000 -c 50
Running 10s test @ http://localhost:3000
50 connections

Stat         Avg       Stdev    Max
Latency (ms) 7.57      2.43     53
Req/Sec      6189.64   768.21   7659
Bytes/Sec    689.62 kB 80.71 kB 851.97 kB

68k requests in 11s, 7.56 MB read

$ autocannon http://localhost:3000 -c 50
Running 10s test @ http://localhost:3000
50 connections

Stat         Avg       Stdev    Max
Latency (ms) 8.05      3.26     58
Req/Sec      5843.82   757.74   7163
Bytes/Sec    647.91 kB 89.57 kB 819.2 kB

64k requests in 11s, 7.13 MB read
```

mcollina:
@marcbachmann 

> Doesn't it really belong into this module? For convenience, yes, that's nice. But in a few cases people might have a custom function. And in that case it's faster to just hook that up on their own.

Yes, you are right.

Let's bite the bullet and make that configurable, with a custom function that wraps it. It should be _inlineable_, so we should not see any real difference in perf.

mcollina:
I've merged this and going to release asap :).

@marcbachmann would you like to help maintaining pino-http?

I look forward to the other PR!

marcbachmann:
> would you like to help maintaining pino-http?

IMO that's fine.
You're just using `npm version [major|minor|patch] && npm publish` to release, right?

mcollina:
@davidmarkclements can you give @marcbachmann access? :)

PR #8, opened 2016-08-18
Fix responseTime on errored request

marcbachmann:
The responseTime on a failed request didn't work anymore


mcollina:
thanks! Well spotted!

PR #9, opened 2016-08-18
Support a custom requestId generation function

marcbachmann:
This adds a `opts.genReqId` option that can be a function that returns a custom request id.

@mcollina Did you have something like this in mind?

Tests are still missing

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,12 +91,25 @@ $ node example.js | pino
 
 ### pinoHttp([options], [stream])
 
-`pino-http` has the same options as [pino](http://npm.im/pino).
+#### options
 
+`pino-http` has the same options as [pino](http://npm.im/pino).
 `pino-http` attaches listeners to the request, in order to log when the request completes
 
+##### logger
+
 `pino-http` can reuse a pino instance if passed with the `logger`
-property:
+property
+
+##### genReqId
+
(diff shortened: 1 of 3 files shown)


mcollina:
Yes I did :)

I'm 👍  with tests and docs.

mcollina:
@marcbachmann ping :). Tests coming?

marcbachmann:
yep. sorry. didn't begin yet. Maybe I have some time next week.

marcbachmann:
@mcollina that one would be ready.
I've removed the old `req.id` behavior: https://github.com/davidmarkclements/pino-http/pull/9/files#diff-7f28009834c7b72b5b0080bc0487ccddL51

So that would be a breaking change. Then again, this wasn't documented.

mcollina:
I'm 👍 for this change.

I would keep the old behavior anyway, if it is not costly (I don't think so).

marcbachmann:
We can also just set `req.log.id` instead of `req.id`. Then that doesn't have any side effects.

mcollina:
@marcbachmann please do!

mcollina:
@marcbachmann is this done?
Sorry, I think we didn't notice this.

marcbachmann:
yep. no problem.

PR #12, opened 2016-09-29
Store the id in a symbol.

### LINKED ISSUE #11: Use a Symbol to store the id of the request.
As titled. So we will not have to override things, etc.

cc @davidmarkclements @marcbachmann

mcollina:
Fixes #11.

@marcbachmann @davidmarkclements what do you think?

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -1,6 +1,7 @@
 'use strict'
 
 var pino = require('pino')
+var id = Symbol('id')
 
 function pinoLogger (opts, stream) {
   if (opts && opts._writableState) {
@@ -41,7 +42,7 @@ function pinoLogger (opts, stream) {
   }
 
   function loggingMiddleware (req, res, next) {
-    req.id = genReqId(req)
+    req[id] = genReqId(req)
     req.log = res.log = logger.child({req: req})
     res.startTime = Date.now()
 
@@ -56,7 +57,7 @@ function pinoLogger (opts, stream) {
 
 function asReqValue (req) {
   return {
-    id: req.id,
+    id: req[id],
     method: req.method,
     url: req.url,
     headers: req.headers,
--- a/logger.js
+++ b/logger.js
@@ -1,7 +1,7 @@
 'use strict'
 
 var pino = require('pino')
-var id = Symbol('id')
+var id = Symbol.for('id')
 
 function pinoLogger (opts, stream) {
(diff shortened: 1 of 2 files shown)


marcbachmann:
Either drop support for node<4 or put it into a try catch. And set a random number. or something like that. ;)

mcollina:
We have symbols in v0.12. That's a bit weird, but symbols have been there for quite a while.

marcbachmann:
oh. I thought that was only available with the harmony flag.
btw. I've always used them with a `_` as prefix to make it more explicit that it's private.
Maybe that's easier to read in some cases. e.g. `req[_id] = genReqId(req)`

mcollina:
I don't even know if this is a good idea or not.

logger.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
 var pino = require('pino')
+var id = Symbol('id')
```
mcollina: that's a good idea in fact.

logger.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
 var pino = require('pino')
+var id = Symbol.for('id')
```
davidmarkclements: should this be Symbol.for('pino-http') - there's a fair chance in the same process someone could do Symbol.for('id') and possibly rely on Symbol.keyFor(sym)

logger.js
```diff
@@ -56,7 +57,7 @@ function pinoLogger (opts, stream) {
 
 function asReqValue (req) {
   return {
-    id: req.id,
+    id: req[id],
```
davidmarkclements: obv this won't stringify so this will affect workflow with log viewers (i.e. kibana) - we index by time anyway, but seeing the unique id on each requests helps to differentiate in your mind

mcollina: this will stringify, we are extracting `req[id]` into the main object.

jsumners:
What problem does this solve? I'm not clear on it.

mcollina:
There are good cases in which we don't want to change the `request`. That's the fix. I don't know if it's worth it or not, really.

davidmarkclements:
Shall we add this to the list of things to discuss in austin?

jsumners:
Might as well go through all the open issues and get thoughts on them.

davidmarkclements:
we all voted to close - austin

PR #16, opened 2016-11-28
Stream options

delvedor:
Added the `stream` option to pass the destination stream inside the `options` object instead of the second parameter.
I left the standalone stream parameter for back compatibility.
(cc https://github.com/mcollina/fastify/pull/25)

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -12,8 +12,9 @@ function pinoLogger (opts, stream) {
   opts.serializers = opts.serializers || {}
   opts.serializers.req = opts.serializers.req || asReqValue
   opts.serializers.res = opts.serializers.res || pino.stdSerializers.res
+  opts.stream = opts.stream || stream
 
-  var logger = wrapChild(opts, opts.stream || stream)
+  var logger = wrapChild(opts, opts.stream)
   var genReqId = reqIdGenFactory(opts.genReqId)
   loggingMiddleware.logger = logger
   return loggingMiddleware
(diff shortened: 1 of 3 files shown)


logger.js
```diff
@@ -13,7 +13,7 @@ function pinoLogger (opts, stream) {
   opts.serializers.req = opts.serializers.req || asReqValue
   opts.serializers.res = opts.serializers.res || pino.stdSerializers.res
 
-  var logger = wrapChild(opts, stream)
+  var logger = wrapChild(opts, opts.stream || stream)
```
marcbachmann: I'm fine with that change, but keep attention, this is passing the stream to the pino instance. That might cause side effects once that api changes.

I'd directly set `opts.stream` instead of using the stream variable fallback so pino always receives the same configuration. No matter whether the stream instance is passed as second arg or on `opts`

```
opts.stream = opts.stream || stream
```

delvedor: Totally agree, I did the change :)

logger.js
```diff
@@ -12,8 +12,9 @@ function pinoLogger (opts, stream) {
   opts.serializers = opts.serializers || {}
   opts.serializers.req = opts.serializers.req || asReqValue
   opts.serializers.res = opts.serializers.res || pino.stdSerializers.res
+  opts.stream = opts.stream || stream
```
mcollina: the problem is that we are always passing `opts.stream`  property inside opts. This will be picked up by pino.

We should do:

```js
var stream = opts.stream || stream
delete opts.stream
```

delvedor: Oops! Sorry, my bad!

davidmarkclements:
We need to update docs for dependants (express pink logger etc)

PR #17, opened 2016-12-01
Add useLevel option

tothandras:
I think it should be using the `debug` level by default, but I didn't want to introduce a breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,27 +89,16 @@ $ node example.js | pino
 
 ## API
 
-### pinoHttp([options], [stream])
+### pinoHttp([opts], [stream])
 
-#### options
+`opts`: it has all the options as [pino](http://npm.im/pino) and
 
-`pino-http` has the same options as [pino](http://npm.im/pino).
-`pino-http` attaches listeners to the request, in order to log when the request completes
+* `logger`: `pino-http` can reuse a pino instance if passed with the `logger` property
+* `genReqId`: you can pass a function which gets used to generate a request id. The first argument is the request itself. As fallback `pino-http` is just using an integer. This default might not be the desired behavior if you're running multiple instances of the app
+* `useLevel`: the logger level `pino-http` is using to log out the response. default: `info`
+* `stream`: same as the second parameter
 
-##### logger
-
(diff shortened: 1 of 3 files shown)


logger.js
```diff
@@ -37,7 +38,7 @@ function pinoLogger (opts, stream) {
       return
     }
 
-    log.info({
+    log[opts.useLevel]({
```
mcollina: can you assign this to a variable?

logger.js
```diff
@@ -12,6 +12,7 @@ function pinoLogger (opts, stream) {
   opts.serializers = opts.serializers || {}
   opts.serializers.req = opts.serializers.req || asReqValue
   opts.serializers.res = opts.serializers.res || pino.stdSerializers.res
+  opts.useLevel = opts.useLevel || 'info'
```
mcollina: we should delete `useLevel`, or it will appear inside the pino log lines.

README.md
```diff
@@ -132,7 +121,9 @@ var logger = require('pino-http')({
   serializers: {
     req: pino.stdSerializers.req,
     res: pino.stdSerializers.res
-  }
+  },
+
+  useLevel: 'debug'
```
mcollina: can you set this to `'info'`, and add a comment that this is the default.

tothandras:
@mcollina updated

test.js
```diff
@@ -83,6 +83,23 @@ test('exposes the internal pino', function (t) {
...
 })
 
+test('uses the log level passed in as an option', function (t) {
+  var dest = split(JSON.parse)
+  var logger = pinoHttp({ useLevel: 'debug', level: 'debug' }, dest)
+
+  function decoratedLogger (req, res, next) {
+    logger(req, res, next)
+    req.log.debug = function () {
+      t.end()
```
mcollina: can you please use the `on('data')` pattern in one of the other test and verify that `useLevel` is not included in the output?

davidmarkclements:
once change req @mcollina is done I'm happy to merge - thanks @tothandras !

test.js
tothandras: I don't see it showing up in the output if the key is not deleted.

PR #18, opened 2016-12-08
Print log for aborted connections

nemisj:
It's possible that requests takes so long that it's either aborted by the browser or by the http.Server.
In both cases, there is no log call saying that the connection is dead, which gives the feeling in the logs, that request is still in progress.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -48,7 +48,7 @@ function pinoLogger (opts, stream) {
 
   function onReqAborted () {
     var res = this.res
-    res.statusCode = 0
+    res.statusCode = 408
     onResFinished.call(res, new Error('Aborted'))
   }
 
--- a/logger.js
+++ b/logger.js
@@ -27,7 +27,6 @@ function pinoLogger (opts, stream) {
   function onResFinished (err) {
     this.removeListener('finish', onResFinished)
     this.removeListener('error', onResFinished)
-    this.req.removeListener('aborted', onReqAborted);
 
     var log = this.log
     var responseTime = Date.now() - this.startTime
@@ -47,9 +46,9 @@ function pinoLogger (opts, stream) {
     }, 'request completed')
   }
 
-  function onReqAborted() {
-    this.res.statusCode = 0;
-    onResFinished.call(this.res, new Error('Aborted'));
+  function onReqAborted (res) {
+    res.statusCode = 0
+    onResFinished.call(res, new Error('Aborted'))
   }
 
   function loggingMiddleware (req, res, next) {
@@ -60,7 +59,9 @@ function pinoLogger (opts, stream) {
     res.on('finish', onResFinished)
     res.on('error', onResFinished)
     // it's possible that browser aborts connection, or http-server because of timeout
-    req.on('aborted', onReqAborted);
+    req.on('aborted', function () {
+      onReqAborted(res)
+    })
 
     if (next) {
       next()
--- a/logger.js
+++ b/logger.js
@@ -27,6 +27,7 @@ function pinoLogger (opts, stream) {
   function onResFinished (err) {
     this.removeListener('finish', onResFinished)
     this.removeListener('error', onResFinished)
+    this.req.removeListener('aborted', onReqAborted);
 
     var log = this.log
     var responseTime = Date.now() - this.startTime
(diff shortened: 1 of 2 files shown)


davidmarkclements:
Yes! @nemisj v. nice

1. Would you mind running comparison benchmark?
2. There's a couple of linting issues

logger.js
```diff
@@ -27,6 +27,7 @@ function pinoLogger (opts, stream) {
   function onResFinished (err) {
     this.removeListener('finish', onResFinished)
     this.removeListener('error', onResFinished)
+    this.req.removeListener('aborted', onReqAborted);
```
davidmarkclements: Can we get away with not removing aborted listener?

If req is done, aborted won't fire and the whole object is heading for gc. I ask because removeListener isn't cheap

marcbachmann: Good idea. Why don't we do the same for the response?

logger.js
```diff
@@ -46,13 +47,20 @@ function pinoLogger (opts, stream) {
     }, 'request completed')
   }
 
+  function onReqAborted() {
+    this.res.statusCode = 0;
```
marcbachmann: @davidmarkclements how do you feel about overwriting the `statusCode` in here?
I wouldn't do it.

nemisj: @marcbachmann, if we don't do it there, then the log will be with 200 statusCode, that's why it's here

nemisj:
@davidmarkclements 

1. how can i do that?
2. don't see any lint config, what rules do you use? what should I change?

davidmarkclements:
@nemisj 

1. linting uses standard (http://npm.im/standard) - run npm test to check
2. yah sorry benchmarking isn't obvious on this module, run `node benchmarks/pino-server.js` and then use [autocannon](http://npm.im/autocannon) to bench the http server, - bench it before the change and after (pref three times each, then take avg) 
3. could you also add tests (see tests.js)

nemisj:
@davidmarkclements, @marcbachmann 
Guys, I've linted, added tests and removed `removeListener`
I haven't tested it yet in a real wordl use-case after deleting 'removeListener', so i will do it to see if it's not leaking and also I will do benchmarking.

logger.js
```diff
@@ -46,13 +46,22 @@ function pinoLogger (opts, stream) {
...
     req.id = genReqId(req)
     req.log = res.log = logger.child({req: req})
     res.startTime = Date.now()
 
     res.on('finish', onResFinished)
     res.on('error', onResFinished)
+    // it's possible that browser aborts connection, or http-server because of timeout
+    req.on('aborted', function () {
+      onReqAborted(res)
+    })
```
mcollina: we should avoid creating this closure if possible, not sure if there is a way. Maybe we can attach the response to the request (if it is not already there)

nemisj:
@davidmarkclements how much connections do you use as default for benchmarking? any other options?

nemisj:
statusCode is not working, indeed. http_server starts to puking if there are some unflushed data.

```
RangeError: Invalid status code: 0
    at ServerResponse.writeHead (_http_server.js:192:11)
```

But 200 is also not nice in the log, any ideas?

mcollina:
IMHO we shouldn't touch it. However, we should detect if the request was aborted, and correctly _not_ printing the status code in the serializer.
@davidmarkclements do you think it could solve your issues?

nemisj:
@mcollina, why do you think we shouldn't touch it? we touch already request object and response object with augmenting it and statusCode in this case is something what can be used as a assertion also outside the logger. I see there is a 408 http code - http://www.checkupdown.com/status/E408.html i guess this is the correct one which should be used instead. What do you think?

mcollina:
@nemisj we shouldn't touch t because the user might change things. But you are right, setting 408 is correct.

nemisj:
@mcollina i will put 408 in the code

logger.js
```diff
@@ -46,13 +46,22 @@ function pinoLogger (opts, stream) {
...
+    var res = this.res
+    res.statusCode = 408
+    onResFinished.call(res, new Error('Aborted'))
+  }
+
   function loggingMiddleware (req, res, next) {
     req.id = genReqId(req)
     req.log = res.log = logger.child({req: req})
     res.startTime = Date.now()
+    if (!req.res) { req.res = res }
```
marcbachmann: I don't get why we have to check for `req.res` before setting it.
IMO we can remove that. Just use `req.res = res` directly.

marcbachmann: i just checked express. They are setting req.res, the node http server isn't.

mcollina: Let's leave as it is then.

PR #19, opened 2016-12-15
Modified genReqId with quick, small performance boost.

gmcnaught:
…estId is faster than modulus by 7x on my Macbook Pro

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -103,7 +103,7 @@ function reqIdGenFactory (func) {
   var maxInt = 2147483647
   var nextReqId = 0
   return function genReqId (req) {
-    return req.id || (nextReqId = (nextReqId % maxInt) + 1)
+    return req.id || (nextReqId = (nextReqId + 1) & maxInt)
   }
 }
 


mcollina:
I'm not seeing any perf boost on pino-http benchmarks. What version of Node are you using?

marcbachmann:
Maybe we can expose the genReqId method on the file so we can test & benchmark it directly.

mcollina:
I mean, I'm fine with the change. I'm more interested if this boost things up here. I'm 100% sure it's faster, I would just like to have a bench attached to the PR before landing.

davidmarkclements:
oh this is really cool - @gmcnaught can you share your node version and OS so we can reproduce your findings - since @mcollina isn't seeing the same thing it might relate to a specific conditions. 

We'll want to check the conditions to see which approach is most future-proof

gmcnaught:
Of course!

First of all, I'm not sure where I pulled 7x out of in my commit message that's just plain wrong - its more like 10% increase in performance for the function, and without knowing how you are benchmarking the full pino-http I can't speak to actual, real world, perf improvement for the whole app.


__Risks__: 
Modulus just works, no matter what the maxInt in genReqId is, whereas using a mask only works if that maxInt is (2^n)-1; [Javascript spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) actually calls for all numbers to be a double, so you could increase maxInt up to (2^52)-1 and still have an extra bit to cause rollover, but if someone just picks a very large number at some point in the future, this will start to fail and it might not be immediately obvious why it is failing.

__Specs__: I'm running a 2015 Macbook Pro, nodejs 6.9.1

Results of just testing genReqID:
```
modulus x 87,768,064 ops/sec ±1.02% (90 runs sampled)
mask x 97,243,797 ops/sec ±1.28% (90 runs sampled)
Fastest is mask by 10.7963%
```

I originally just compared the time elapsed with process.hrtime() between running both 10,000x but the below code seemed better suited for a PR

The actual benchmark code I used to generate these results:
```
var Benchmark = require('benchmark');

var req = {};
var maxInt = 2147483647;
var nextReqId = 0;
var suite = new Benchmark.Suite;

suite.add('modulus',function(){
  return req.id || (nextReqId = (nextReqId % maxInt) + 1)
});

suite.add('mask',function(){
  return req.id || (nextReqId = (nextReqId + 1) & maxInt)
});

// add listeners
suite.on('cycle', function(event) {
  console.log(String(event.target));
});
suite.on('complete', function() {
  //console.log(this);
   var fastest = this.filter('fastest').map('hz');
  var slowest = this.filter('slowest').map('hz');

  var diff = ((fastest - slowest)/slowest * 100).toFixed(4);
  console.log('Fastest is ' + this.filter('fastest').map('name') + ' by '+diff+'%');

});

suite.run();
```

mcollina:
awesome work. Is it faster in practice over http? How did you find out that this was a problem? I'm trying to bench over HTTP, and I get very similar results before and after.

davidmarkclements:
@gmcnaught this is really good info, the way we bench pino-http isn't super obvious (we should doc this as it's not the first time that question has come up) - basically run benchmarks/pino-server and use http://npm.im/autocannon (defaults are fine) to load test

gmcnaught:
I would say this change would fall under "premature optimizations born out of curiosity".

When I saw that request IDs were integers, I looked in pino-http to verify it did indeed roll-over at some point, so I knew before I put something in production I wouldn't start getting negative request IDs... and then just tried using bitmask instead of 

The original method at ~11 nanoseconds per operation and the new method at ~10 nanoseconds per operation.. 1 nanosecond per operation in an http request with a total of 10ms isn't going to even show up on any meaningful graph of improvement :)

mcollina:
Awesome, I *love* it.

PR #21, opened 2017-01-22
Don't log aborted on GET and HEAD

ronag:


mcollina:
linting is failing, see https://travis-ci.org/pinojs/pino-http/jobs/194234592#L605.

mcollina:
Why did you close this? I think it still need fixing.

logger.js
```diff
@@ -36,20 +36,23 @@ function pinoLogger (opts, stream) {
...
 
   function onReqAborted () {
     var res = this.res
-    res.statusCode = 408
-    onResFinished.call(res, new Error('Aborted'))
+    if (req.method !== 'HEAD' && req.method !== 'GET') {
+      res.statusCode = 408
+      onResFinished.call(res, new Error('aborted'), 'aborted')
+    } else {
+      onResFinished.call(res, undefined, 'aborted')
```
davidmarkclements: should be null rather than undefined here

PR #22, opened 2017-01-26
Handle client abort and server timeout separatly.

ronag:


logger.js
```diff
@@ -62,6 +78,7 @@ function pinoLogger (opts, stream) {
     res.on('error', onResFinished)
     // it's possible that browser aborts connection, or http-server because of timeout
     req.on('aborted', onReqAborted)
+    req.socket.on('timeout', onReqTimeout.bind(req))
```
ronag: Not sure why `req.on('timeout', ...)` doesn't do the same thing...

test.js
```diff
@@ -291,9 +291,8 @@ test('react on aborted event from clientside', function (t) {
   }, logger)
 
   dest.on('data', function (line) {
-    t.equal(line.msg, 'request errored', 'message is set')
-    t.equal(line.err.message, 'Aborted', 'error message is set')
-    t.equal(line.res.statusCode, 408, 'statusCode is 408')
+    t.equal(line.msg, 'request aborted', 'message is set')
+    // t.equal(line.err.message, 'Aborted', 'error message is set')
```
ronag: Need separate test for safe and non-safe methods.

logger.js
```diff
@@ -24,7 +24,7 @@ function pinoLogger (opts, stream) {
   loggingMiddleware.logger = logger
   return loggingMiddleware
 
-  function onResFinished (err) {
+  function onResFinished (err, msg = 'completed') {
```
mcollina: `pino-http` supports node v4+, and this syntax is available only on node v6+. Moreover, it is slower than using `||` to set a default.

logger.js
mcollina: I think `'timeout'`  is not emitted if `setTimeout()`  is not called on the request or response.
Maybe is what we want, would you mind setting `setTimeout` explicitly?

`bind` is extremely slow on node v4 and v6, it should not be used.

ronag: I don't think it's a good idea to override the timeout.

mcollina: I agree, but that's why `req.on('timeout')` is not firing.

ronag:
@mcollina Updated

mcollina:
Released as v2.4.0.

PR #26, opened 2017-02-16
Client abort is not an error.

ronag:
Hopefully without memory leak this time.


logger.js
```diff
@@ -58,10 +62,17 @@ function pinoLogger (opts, stream) {
...
+    res.once('error', onResFinished)
+
+    req.once('close', onReqClosed)
+    req.once('aborted', onReqAborted)
+    req.onSocketTimeout = function () {
+      req.removeListener('aborted', onReqAborted)
+      res.statusCode = 408
+      onResFinished.call(res, new Error('Timeout'))
+    }
+    req.socket.once('timeout', req.onSocketTimeout)
```
mcollina: I'm not convinced we should listen to `socket.on('timeout')` in the first place.

ronag: @mcollina: Do you have a suggestion?

ronag: @mcollina: We could listen to `res.on('timeout', ...)` as well. The problem is that it will actually change the behavior of the associated socket.

ronag: Node uses `socket.on('timeout', ...)` to implement `res.on('timeout')` as well as `req.on('timeout')`. https://github.com/nodejs/node/blob/a4bb9fdb893c8f2c36a6b77862e29b2608e080c2/lib/_http_server.js#L285

mcollina: The problem is that adding that listener to a socket can cause massive issues with http pipelining, and this is why we had to revert the previous fix.

it does not seem https://github.com/nodejs/node/blob/a4bb9fdb893c8f2c36a6b77862e29b2608e080c2/lib/_http_server.js#L374 changes the behavior of the socket. Can you explain further?

ronag: What happens is that if you put a timeout listener on either res or req the socket will not be destroyed by itself. Sure, we could destroy the socket on timeout, but then what if someone wants that behavior?

ronag: > to a socket can cause massive issues with http pipelining

Please explain? Adding timeout listener on the socket is exactly what node does internally already? https://github.com/nodejs/node/blob/a4bb9fdb893c8f2c36a6b77862e29b2608e080c2/lib/_http_server.js#L285

davidmarkclements:
can part of this PR include a test for the memleak - test should fail when applied to previous attempt, and not fail when applied to this PR - we can use `v8.getHeapSpaceStatistics`

doesn't have to be a unit test, just a little bit of rigging to apply pressure and get memory growth over time

ronag:
@davidmarkclements: Not sure how to do that...

logger.js
mcollina: The leak was caused by the fact that the `'timeout'` listener was not properly removed.

davidmarkclements:
Take a look at http://npm.im/climem, see how it gathers numbers with the core v8 module (`require('v8')`) and how you inject it into process arbitrarily with -r flag

Use http://npm.im/autocannon to stress test

Even without a formal test these are the tools we'd be using to check for memory leaks `node --inspect` will also be helpful for general profiling

mcollina:
I'd rather *not* handle the timeout event (at all), in `pino-http`, as it seems this use case is not properly supported within node `'http'` module.

ronag:
@mcollina: The thing a want to avoid is, when using e.g. hls.js and the client aborts (correctly) requests I don't want to spam my log with aborted errors...

mcollina:
can we just not log those as errors, maybe as warnings, or just info?
Or maybe put the listener to `'abort'` behind a flag?

ronag:
@mcollina: I'm fine with that.

mcollina:
@davidmarkclements @jsumners @marcbachmann ?

ronag:
@mcollina: How about now?

mcollina:
I'm definitely 👍  with that.

jsumners:
I don't really have time to dig into the details of this right now. So I'm going to abstain from providing an opinion.

ronag:
@mcollina ping on this?

mcollina:
Merging, as no one else has anything against this. Let's see if this goes through.

PR #28, opened 2017-04-10
use standard err serializer

ahdinosaur:
hey,

this pull request adds `pino.stdSerializers.err` to the logger serializers.

is there a reason why this was not done before? maybe i'm naive and missing something obvious.

cheers!


mcollina:
I left that to the user to add, but I think that's a very good point that _everybody_ is going to need it.
So, I'm 👍  for this.

LGTM

@marcbachmann @davidmarkclements what do you think?

PR #33, opened 2017-06-12
Fix code example in readme

peyerluk:
The stdSerializers are defined in `pinoHttp` and this was mixed up with `pino` in one code example.


peyerluk:
Closing again. I had an instantiated pino logger where `stdSerializers `are not defined. That's what confused me. I saw that the `stdSerializers` are defined both on `pino` and `pinoHttp`.

PR #44, opened 2017-12-09
Fix setting request id

jsumners:
While updating `restify-pino-logger` I learned that the request id property was not getting set properly. I'm don't know how it ever was. Consider:

1. https://github.com/pinojs/pino-http/blame/626eede60281e460e77968836293d671d709f19a/logger.js#L53
1. https://github.com/pinojs/pino-http/blame/626eede60281e460e77968836293d671d709f19a/logger.js#L82

First it's set to a function and then that function is assigned to the property.

This PR simply invokes the `req.id` function in the serializer if it is actually a function. Otherwise it assigns whatever that value is to the property.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -121,7 +121,7 @@ function wrapReqSerializer (serializer) {
 function asReqValue (req) {
   var connection = req.connection
   const _req = Object.create(pinoReqProto)
-  _req.id = Function.prototype.isPrototypeOf(req.id) ? req.id() : req.id
+  _req.id = typeof req.id === 'function' ? req.id() : req.id
   _req.method = req.method
   _req.url = req.url
   _req.headers = req.headers
--- a/logger.js
+++ b/logger.js
@@ -121,7 +121,7 @@ function wrapReqSerializer (serializer) {
 function asReqValue (req) {
   var connection = req.connection
   const _req = Object.create(pinoReqProto)
-  _req.id = req.id
+  _req.id = Function.prototype.isPrototypeOf(req.id) ? req.id() : req.id
   _req.method = req.method
   _req.url = req.url
   _req.headers = req.headers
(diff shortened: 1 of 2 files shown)


logger.js
```diff
@@ -121,7 +121,7 @@ function wrapReqSerializer (serializer) {
 function asReqValue (req) {
   var connection = req.connection
   const _req = Object.create(pinoReqProto)
-  _req.id = req.id
+  _req.id = Function.prototype.isPrototypeOf(req.id) ? req.id() : req.id
```
mcollina: why not `typeof req.id === 'function'`? Which one is faster?

davidmarkclements: unless there's a speed benefit.. definitely go with `typeof` here...

test.js
```diff
@@ -414,3 +414,29 @@ test('res.raw is not enumerable', function (t) {
...
 })
+
+test('req.id has a non-function value', function (t) {
+  t.plan(1)
+  var dest = split(JSON.parse)
+  var logger = pinoHttp({
+    logger: pino(dest),
+    serializers: {
+      req: function (req) {
+        t.is(Function.isPrototypeOf(req.id), false)
```
davidmarkclements: prefer `typeof` - unless there's some really awesome reasons

test.js
```diff
@@ -414,3 +414,29 @@ test('res.raw is not enumerable', function (t) {
...
+        return req
+      }
+    }
+  })
+
+  var server = http.createServer(handler)
+  server.unref()
+  server.listen(0, () => {
+    const port = server.address().port
+    http.get(`http://127.0.0.1:${port}`, () => {})
```
davidmarkclements: pointless pedantry: you could just do `http.get(server.address(), () => {})`

logger.js
jsumners: /shrug, I prefer using the actual API. I haven't tested for any difference in speed.

PR #45, opened 2018-01-25
Fix request on error logger

natanavra:
The `err` param in `onResFinished(err)` was always `undefined` no matter what the result of the request was.

From nodejs / express documentation, there is no event named `error`:
- https://expressjs.com/en/4x/api.html#res
- https://nodejs.org/api/http.html#http_class_http_serverresponse

The `res` object can be used as a the `err` holder, and set on the `res` object in the express error handler middleware 
```js
function onErrorMiddleware(err, req, res, next) {
    res.err = err;
    //code
}
```

If no `err` object on the response but a bad `statusCode`, we generate the error from the logger. I know this would show a somewhat confusing stack, but it's something. 
Besides the `stdSerializers.err` handles string error as an array of chars.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -34,7 +34,7 @@ function pinoLogger (opts, stream) {
     var log = this.log
     var responseTime = Date.now() - this[startTime]
 
-    if (err || this.err || this.statusCode >= 400) {
+    if (err || this.err || this.statusCode >= 500) {
       log.error({
         res: this,
         err: err || this.err || new Error('failed with status code' + this.statusCode),
--- a/logger.js
+++ b/logger.js
@@ -27,16 +27,17 @@ function pinoLogger (opts, stream) {
   loggingMiddleware.logger = logger
   return loggingMiddleware
 
-  function onResFinished () {
+  function onResFinished (err) {
+    this.removeListener('error', onResFinished)
     this.removeListener('finish', onResFinished)
 
     var log = this.log
     var responseTime = Date.now() - this[startTime]
 
-    if (this.err || this.statusCode >= 400) {
+    if (err || this.err || this.statusCode >= 400) {
       log.error({
         res: this,
-        err: this.err || new Error('failed with status code' + this.statusCode),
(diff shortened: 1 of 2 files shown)


mcollina:
Thinking about it some more, I think we have to support both things, as other frameworks might emit errors as `'error'` events, can you please maintain the previous functionality while adding this one?

natanavra:
I was pretty sure this is an express only package, since I came here from [express-pino-logger](https://github.com/pinojs/express-pino-logger)

logger.js
```diff
@@ -28,16 +28,16 @@ function pinoLogger (opts, stream) {
...
   function onResFinished (err) {
-    this.removeListener('finish', onResFinished)
     this.removeListener('error', onResFinished)
+    this.removeListener('finish', onResFinished)
 
     var log = this.log
     var responseTime = Date.now() - this[startTime]
 
-    if (err) {
+    if (err || this.err || this.statusCode >= 400) {
```
mcollina: I think this should be `>= 500`, not `400`.

natanavra: I think that you'd want to have logs of issues like: bad request, unauthorized, forbidden access

And besides, we want to log the request / response, and a 4xx is a request (client) error.

mcollina: Non of them are errors.
This is a flow for error conditions.

@jsumners what do you think?

natanavra: One more point on my side, if I'd look at a logs dashboard, I'd be looking to see 4xx response, these can indicate a broken client side library, or a broken flow somewhere.

natanavra:
Basically, we can leave it up to the user to set the `err` on the `res`, that would probably still result in 4xx errors being logged, as I guess most developers would call `next(err4xx);`.

mcollina:
I think so, yes.

natanavra:
Modified. Not sure it notified you, as the commit and your message happened pretty much the same time... so I'm sending a message.

logger.js
jsumners: Looks like it got resolved before I came around. I'm fine with the current solution.

PR #49, opened 2018-04-06
Fix message formatting in request error fallback

marcbachmann:
The fallback error message didn't contain a space:
`"err":{"type":"Error","message":"failed with status code500",`

The updated message now looks like:
`"err":{"type":"Error","message":"failed with status code 500",`


mcollina:
Can you add a test while you are at it? So we won't regress.

PR #50, opened 2018-04-30
Better formatting for default error

WickyNilliams:
Would previously log a message like `failed with status code500`, this PR simply adds a space between the status code and the supporting text: `failed with status code 500`


davidmarkclements:
awesome, thanks @WickyNilliams

PR #56, opened 2018-11-04
Add getUseLevel option

EladHeller:
Adding a getUseLevel function that dynamically allows you to set the log level based on the status of the result and the error object (if any)

The change makes it possible to separate http errors from system errors.
For example, you can specify a user who types in an incorrect password will get an error with an HTTP 401 status, but the log level will be a warning rather than an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,8 @@ $ node example.js | pino
 * `logger`: `pino-http` can reuse a pino instance if passed with the `logger` property
 * `genReqId`: you can pass a function which gets used to generate a request id. The first argument is the request itself. As fallback `pino-http` is just using an integer. This default might not be the desired behavior if you're running multiple instances of the app
 * `useLevel`: the logger level `pino-http` is using to log out the response. default: `info`
-* `customLogLevel`: you can pass a function which gets used to determine the logger level. The first argument is the response. The second argument is the error object if it passed. If you pass this option don't pass `useLevel` option.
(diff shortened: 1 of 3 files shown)


mcollina:
@jsumners can you take a look as well?

README.md
```diff
@@ -96,6 +96,7 @@ $ node example.js | pino
 * `logger`: `pino-http` can reuse a pino instance if passed with the `logger` property
 * `genReqId`: you can pass a function which gets used to generate a request id. The first argument is the request itself. As fallback `pino-http` is just using an integer. This default might not be the desired behavior if you're running multiple instances of the app
 * `useLevel`: the logger level `pino-http` is using to log out the response. default: `info`
+* `customLogLevel`: you can pass a function which gets used to determine the logger level. The first argument is the response. The second argument is the error object if it passed. If you pass this option don't pass `useLevel` option.
```
jsumners: Hmm, @mcollina do you think this is better named something like `levelHook` or `levelInterceptor`?

Either way, the text should be modified to:

> set to a function `(res, err) => { /* returns level name string */ }`. This function will be invoked to determine the level at which the log should be issued. This option is mutually exclusive with the `useLevel` option. The first argument is the HTTP response. The second argument is an error object if an error has occurred in the request.

README.md
```diff
@@ -124,7 +125,18 @@ var logger = require('pino-http')({
   },
 
   // Logger level is `info` by default
-  useLevel: 'info'
+  useLevel: 'info',
+
+
+  // Define a custom logger level
+  customLogLevel: function (err, res) {
```
jsumners: Signature is backward. Need to swap `err` and `res`.

PR #57, opened 2018-11-09
Update README.md

vincentchauhk01:


mcollina:
Why this change?

vincentchauhk01:
`require('pino')().stdSerializers` is undefined, while `require('pino').stdSerializers` works

PR #62, opened 2019-04-05
add support for wrapping the default error serializer

jstewmon:
Tests will fail initially b/c depends on https://github.com/pinojs/pino-std-serializers/pull/18. Will need to bump version of pino-std-serializers.


jstewmon:
Working as expected with pino-std-serializers@^2.4.0 :-)

PR #67, opened 2019-06-13
feat: add support for request-received middleware (closes #66)

### LINKED ISSUE #66: Add support for request-received middleware
The `request-received` middleware at https://github.com/cabinjs/request-received exposes a `Symbol.for` property, which this package could conditionally consume if interested (as opposed to just generating a `new Date()`.

e.g. https://github.com/pinojs/pino-http/blob/master/logger.js#L63 would change as follows:

```diff
-    res[startTime] = res[startTime] || Date.now()
+    res[startTime] = res[startTime] || req[Symbol.for('request-received.startTime')] ? req[Symbol.for('request-received.startTime').getTime() : Date.now()
```

Ref: #65

niftylettuce:


logger.js
```diff
@@ -4,6 +4,7 @@ var pino = require('pino')
 var serializers = require('pino-std-serializers')
 
 var startTime = Symbol('startTime')
```
davidmarkclements:
```suggestion
```

logger.js
```diff
@@ -60,7 +61,7 @@ function pinoLogger (opts, stream) {
   function loggingMiddleware (req, res, next) {
     req.id = genReqId(req)
     req.log = res.log = logger.child({req: req})
-    res[startTime] = res[startTime] || Date.now()
+    res[startTime] = res[startTime] || req[requestReceivedStartTime] ? req[requestReceivedStartTime].getTime() : Date.now()
```
davidmarkclements:
```suggestion
    res[startTime] = res[startTime] || Date.now()
```

niftylettuce:
@davidmarkclements 	these aren't symbols being exposed by pino, they're exposed by [request-received](https://github.com/cabinjs/request-received), which I created as a standard for Koa and Express communities

davidmarkclements:
cool, since you pushed it two hours ago and don't have anyone depending on it it'll be super easy to update. If one module consumes another, the consuming module integrates using consumed modules API's - rarely the other way round.

niftylettuce:
@davidmarkclements Would you be open to moving https://github.com/cabinjs/request-received to pino organization and letting me have access to it?  If so I will rename these 😄

niftylettuce:
The point of the package is to simply expose these as Symbols for any package to consume - I didn't want to prefix it with `pino-http` since it's not related to `pino-http`, it simply adds to the request object a `process.hrtime()` and `new Date()` object using the Symbol.

niftylettuce:
I guess I could however update the package so that it also defines `req[pinoHttpStartTimeSymbol]` in addition.

davidmarkclements:
I get what you're trying to do with your package but this isn't the way to do, you'll have to think of another way. Closing - feel free to reopen if it becomes useful to you in the requested form

davidmarkclements:
sorry I didn't see https://github.com/pinojs/pino-http/pull/67#issuecomment-501911317 before closing - that could work.

niftylettuce:
The purpose of this isn't so people use Cabin, it's so people that don't want to use Cabin's middleware and simply want to use `parse-request` can do so.

https://github.com/cabinjs/parse-request (major update coming soon)

davidmarkclements:
feel free to reopen if plan is to consume the pino symbol

niftylettuce:
def will do

PR #71, opened 2019-09-10
Upgrade pino-std-serializers to avoid DeprecationWarnings

camlegleiter:
Pull in the latest version of pino-std-serializers to update how the response
headers are retrieved when logging and avoid getting DeprecationWarnings in
Node 12.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "logger.js",
   "dependencies": {
     "pino": "^5.0.0",
-    "pino-std-serializers": "^2.4.0"
+    "pino-std-serializers": "^2.4.2"
   },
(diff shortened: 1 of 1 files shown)


mcollina:
This should be picked up automatically by npm. Maybe you forgot to run `npm update`? Have you tried removing your lock files and `node_modules` and reinstalling?

camlegleiter:
>This should be picked up automatically by npm

This actually led me on a rabbit chase through our dependencies, and it turns out there was actually a different module causing the DeprecationWarnings to pop up. This isn't an issue anymore, so I'll just close this. Thanks for the help!

Issue #24, opened 2017-02-02
timeouts when using pino-http 2.4 + express under alpine linux


stevejhiggs:
Apologies for the brevity of this, I'll add more information as I get it but I thought it was a sufficiently worrying issue to raise straightaway.

Setup:

Using pino-http under express running on the alpine-node:7.4.0 docker image in aws.

hit the express endpoint with a lot of traffic and everything starts backing up + memory usage grows continually. Eventually the server just starts returning timeout errors.

Under 2.3 everything works fine.

I'll add more info as my investigation continues.

mcollina:
This is the culprit then: https://github.com/pinojs/pino-http/pull/22.

It should not be related to alpine, and you should be able to reproduce it everywhere if that's the case.

Does reverting back to 2.3 fix the issue?

stevejhiggs:
yep, 2.3 works fine when I downgraded the package.

It caught us out because we were actually using the express-pino-logger package, which then pulled in pino-http so our deps changed even though the versions were locked. 

We've changed to using pino-http directly now and locked in to 2.3. We're also going to switch to yarn to stop sub-deps upgrading on us :)

mcollina:
Ok, that's definitely the bug then.

cc @ronag

ronag:
@mcollina: Tried to fix: https://github.com/pinojs/pino-http/pull/26

PR #75, opened 2019-12-10
Fix #23 use getHeaders() instead of deprecated _headers (#24)

wibobm:
Bump pino-std-serializers patch version dependency to apply the fix to avoid the deprecation _headers warning.


mcollina:
I'm a bit lost on the need for this. Your npm install should pick this up automatically. Try running `npm update --depth 999` or removing `package-lock.json` and `node_modules`.

PR #76, opened 2020-01-10
Add log request body example

naseemkullah:
Adds an example for a common use case of wanting the request body logged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ var logger = require('pino-http')({
 
 Logging of requests' bodies is disabled by default since it can cause security risks such as having private user information (password, other GDPR-protected data, etc.) logged (and persisted in most setups). However if enabled, sensitive information can be redacted as per [redaction documentation](http://getpino.io/#/docs/redaction).
 
-Furthermore, logging more bytes does slow down throughput ([although the cost is negligible in most use cases](https://www.eschrade.com/page/the-cost-of-logging/)).
+Furthermore, logging more bytes does slow down throughput. [This video by pino maintainers Matteo Collina & David Mark Clements](https://www.youtube.com/watch?v=zja-_IYNrFc&feature=youtu.be) goes into this in more detail.
 
 After considering these factors, logging of the request body can be adchieved as follows:
 
--- a/README.md
+++ b/README.md
@@ -248,7 +248,11 @@ var logger = require('pino-http')({
 })
 ```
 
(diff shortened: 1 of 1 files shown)


naseemkullah:
cc @mcollina

naseemkullah:
Updated as such.

naseemkullah:
Yes and yes.

I am a little unfamiliar about the performance consequences apart from spitting out more bytes in theory reduces performance. 
I mean if the body is huge, I could see why. Is there a way to measure this accurately? e.g. How much does an extra character slow down a request?

Showing how to redact in case you want to log the body but there is sensitive data? This would be great but a little beyond me, could you provide a snippet?

mcollina:
Yes, including more bytes slows down throughput. I don't have any specific data, because it depends on where you are sending the data to, and how fast that destination is.

As for log redaction, you can find it documented at http://getpino.io/#/docs/redaction.

naseemkullah:
Changes applied.

README.md
```diff
@@ -248,6 +248,24 @@ var logger = require('pino-http')({
 })
 ```
 
+Logging of requests' bodies is disabled by default since it can cause security risks such as having private user information (password, other GDPR-protected data, etc.) logged (and persisted in most setups). However if enabled, sensitive information can be redacted as per [redaction documentation](http://getpino.io/#/docs/redaction).
+
+Furthermore, logging more bytes does slow down throughput ([although the cost is negligible in most use cases](https://www.eschrade.com/page/the-cost-of-logging/)).
+
+After considering these factors, logging of the request body can be adchieved as follows:
```
mcollina:
```suggestion
After considering these factors, logging of the request body can be achieved as follows:
```

README.md
```diff
@@ -248,6 +248,24 @@ var logger = require('pino-http')({
 })
 ```
 
+Logging of requests' bodies is disabled by default since it can cause security risks such as having private user information (password, other GDPR-protected data, etc.) logged (and persisted in most setups). However if enabled, sensitive information can be redacted as per [redaction documentation](http://getpino.io/#/docs/redaction).
+
+Furthermore, logging more bytes does slow down throughput ([although the cost is negligible in most use cases](https://www.eschrade.com/page/the-cost-of-logging/)).
```
mcollina: Please remove this article link. I have strong evidence that says otherwise, at least for Node.js apps.

https://youtu.be/zja-_IYNrFc

cc @davidmarkclements

naseemkullah: va bene

mcollina:
Good work, thanks!

naseemkullah:
My pleasure!

naseemkullah:
My pleasure!

blairliu509:
How to log response body?

PR #77, opened 2020-01-27
Add autologgingOptions to ignore certain paths like those for health checks.

katlimruiz:
I would like to push these changes to allow the library to ignore certain paths to autolog.

This is specially useful if I have a health check path which gets called every 3 seconds. If not ignored this would fill our logs a lot.

My first step is to push this in pino-http, then update koa-pino-logger, and I finally use it on my koa app.

One line that definitely needs feedback is on middleware, since the variable is currently called `_currentUrl` (it can improve) and also I did not have access to request on `onResFinished`.

Also I did not find access to the url's path alone, so I had to use raw url for now.

Well hope this can help and be accepted. I think it delivers a good functionality.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -36,7 +36,8 @@ function pinoLogger (opts, stream) {
   var autoLogging = (opts.autoLogging !== false)
   delete opts.autoLogging
 
-  var autoLoggingOptions = opts.autoLoggingOptions
+  var autoLoggingIgnorePaths = opts.autoLoggingOptions.ignorePaths
+  delete opts.autoLoggingOptions.ignorePaths
   delete opts.autoLoggingOptions
 
   var logger = wrapChild(opts, theStream)
@@ -62,8 +63,10 @@ function pinoLogger (opts, stream) {
     }
 
     var shouldLogSuccess = true
-    if (this._currentUrl && autoLoggingOptions && autoLoggingOptions.ignorePaths && autoLoggingOptions.ignorePaths.length) {
-      shouldLogSuccess = autoLoggingOptions.ignorePaths.find(x => this._currentUrl.match(x)) == null
+    if (this._currentUrl && autoLoggingIgnorePaths.length) {
+      for (let i = 0; shouldLogSuccess && i < autoLoggingIgnorePaths.length; i++) {
+        shouldLogSuccess = this._currentUrl.match(autoLoggingIgnorePaths[i]) == null
+      }
     }
 
     if (shouldLogSuccess) {
--- a/logger.js
+++ b/logger.js
@@ -77,9 +77,9 @@ function pinoLogger (opts, stream) {
 
     if (autoLogging) {
       if (req.url && autoLoggingIgnorePaths.length) {
-        for (let i = 0; shouldLogSuccess && i < autoLoggingIgnorePaths.length; i++) {
-          shouldLogSuccess = req.url.match(autoLoggingIgnorePaths[i]) == null
-        }
+        // force the base, because req.url will not have it and URL class needs it.
+        var url = new URL(req.url, 'http://localhost')
+        shouldLogSuccess = !autoLoggingIgnorePaths.includes(url.pathname)
       }
 
       if (shouldLogSuccess) {
--- a/logger.js
+++ b/logger.js
@@ -69,12 +69,13 @@ function pinoLogger (opts, stream) {
   }
 
   function loggingMiddleware (req, res, next) {
+    var shouldLogSuccess = true
+
     req.id = genReqId(req)
     req.log = res.log = logger.child({req: req})
     res[startTime] = res[startTime] || Date.now()
 
     if (autoLogging) {
-      var shouldLogSuccess = true
       if (req.url && autoLoggingIgnorePaths.length) {
         for (let i = 0; shouldLogSuccess && i < autoLoggingIgnorePaths.length; i++) {
           shouldLogSuccess = req.url.match(autoLoggingIgnorePaths[i]) == null
--- a/logger.js
+++ b/logger.js
@@ -62,12 +62,10 @@ function pinoLogger (opts, stream) {
       return
     }
 
-    if (this.shouldLogSuccess) {
-      log[level]({
-        res: this,
-        responseTime: responseTime
-      }, 'request completed')
-    }
+    log[level]({
+      res: this,
+      responseTime: responseTime
+    }, 'request completed')
   }
 
   function loggingMiddleware (req, res, next) {
@@ -85,8 +83,10 @@ function pinoLogger (opts, stream) {
         }
       }
 
-      res.shouldLogSuccess = shouldLogSuccess
-      res.on('finish', onResFinished)
+      if (shouldLogSuccess) {
+        res.on('finish', onResFinished)
+      }
+
       res.on('error', onResFinished)
     }
 
--- a/logger.js
+++ b/logger.js
@@ -2,8 +2,7 @@
 
 var pino = require('pino')
 var serializers = require('pino-std-serializers')
-var URL = require('url').URL
-
+var URL = require('fast-url-parser')
 var startTime = Symbol('startTime')
 
 function pinoLogger (opts, stream) {
@@ -78,8 +77,7 @@ function pinoLogger (opts, stream) {
 
     if (autoLogging) {
       if (req.url && autoLoggingIgnorePaths.length) {
-        // force the base, because req.url will not have it and URL class needs it.
-        var url = new URL(req.url, 'http://localhost')
+        var url = URL.parse(req.url)
         shouldLogSuccess = !autoLoggingIgnorePaths.includes(url.pathname)
       }
 
--- a/logger.js
+++ b/logger.js
@@ -30,7 +30,7 @@ function pinoLogger (opts, stream) {
   delete opts.stream
 
   var autoLogging = (opts.autoLogging !== false)
-  var autoLoggingIgnorePaths = ((opts.autoLogging || {}).ignorePaths || [])
+  var autoLoggingIgnorePaths = (opts.autoLogging && opts.autoLogging.ignorePaths) ? opts.autoLogging.ignorePaths : []
   delete opts.autoLogging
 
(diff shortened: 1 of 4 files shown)


logger.js
```diff
@@ -54,17 +61,27 @@ function pinoLogger (opts, stream) {
       return
     }
 
-    log[level]({
-      res: this,
-      responseTime: responseTime
-    }, 'request completed')
+    var shouldLogSuccess = true
+    if (this._currentUrl && autoLoggingOptions && autoLoggingOptions.ignorePaths && autoLoggingOptions.ignorePaths.length) {
+      shouldLogSuccess = autoLoggingOptions.ignorePaths.find(x => this._currentUrl.match(x)) == null
```
mcollina: Would you mind to:

1. extract `ignorePaths` into its own variable
2. remove the use of `find()` and use a for(;;) loop instead? `find()` is significantly slower.

logger.js
```diff
@@ -54,17 +61,27 @@ function pinoLogger (opts, stream) {
       return
     }
 
-    log[level]({
-      res: this,
-      responseTime: responseTime
-    }, 'request completed')
+    var shouldLogSuccess = true
+    if (this._currentUrl && autoLoggingOptions && autoLoggingOptions.ignorePaths && autoLoggingOptions.ignorePaths.length) {
```
mcollina: There is no need to verify _all_ of these for every requests, as most should not change.

logger.js
```diff
@@ -54,17 +61,27 @@ function pinoLogger (opts, stream) {
...
+      }, 'request completed')
+    }
   }
 
   function loggingMiddleware (req, res, next) {
     req.id = genReqId(req)
     req.log = res.log = logger.child({req: req})
     res[startTime] = res[startTime] || Date.now()
-    if (!req.res) { req.res = res }
+    res._currentUrl = req.url
```
mcollina: I would really _not_ store this here. I would rather run the "should log" check here and change the log level to the child logger to `'error'` or even `'silent'`.

katlimruiz: I will send the change, for your review. However, I do have one note here. I wanted to keep the current structure where the events (finish, error) get sent anyway because if in the future those event listeners need to execute other logic, then it would not be ran. That is why I wanted to ignore the path at the last moment.

katlimruiz: But yes, what I can do is to calculate the flag in the middleware, and then just use it in the finish event.

katlimruiz: On the other hand, how would you go with `silent` level and customLogLevel?

katlimruiz: about the silent level, I tried this:

```
if (!this.shouldLogSuccess) {
  level = 'silent'
}
log[level](...)
```

and it didn't work. Log object does not have a silent function (and I think in [2018 ](https://github.com/pinojs/pino/issues/434) it was deleted).

logger.js
katlimruiz: fixed

logger.js
katlimruiz: fixed

logger.js
katlimruiz: fixed. waiting for your comments.

logger.js
```diff
@@ -54,19 +62,30 @@ function pinoLogger (opts, stream) {
...
+      // this is to ensure that the event listener still executes (as there may be other logic that is needed).
+      var shouldLogSuccess = true
+      if (req.url && autoLoggingIgnorePaths.length) {
+        for (let i = 0; shouldLogSuccess && i < autoLoggingIgnorePaths.length; i++) {
+          shouldLogSuccess = req.url.match(autoLoggingIgnorePaths[i]) == null
+        }
+      }
+
+      res.shouldLogSuccess = shouldLogSuccess
       res.on('finish', onResFinished)
```
mcollina: Instead of keeping some state around, why can't you just avoid to add this listener?

katlimruiz: Ok, i can do that. I was explaining that I didn't want to cut off future logic in that listener, but thats ok. let me push again.

mcollina: I think it would have less impact on the overhead if we avoid to add that listener.

katlimruiz: fixed

katlimruiz: however, do note that the listener will still execute the this.removeListener('finished') even when that listener was not registered. It doesn't fail, however it doesn't seem quite right.

So I divided onResError and onResFinished, but refactoring further did not work as expected. When I tried refactoring into `getResponseTime` and `getLogLevel`, and passing the `this` (response), one test failed ('err.raw is available to custom serializers').

logger.js
```diff
@@ -64,10 +72,21 @@ function pinoLogger (opts, stream) {
     req.id = genReqId(req)
     req.log = res.log = logger.child({req: req})
     res[startTime] = res[startTime] || Date.now()
-    if (!req.res) { req.res = res }
 
     if (autoLogging) {
-      res.on('finish', onResFinished)
+      // calculate here whether to log success, but apply this flag up to the last minute.
+      // this is to ensure that the event listener still executes (as there may be other logic that is needed).
+      var shouldLogSuccess = true
```
mcollina: Can you move this to the top? It's weird to have a variable definition within an if that it's used outside of scope.

logger.js
mcollina: I think this approach is correct, don't mind for the double removeListener, it's very fast anyway.

logger.js
katlimruiz: even if autologging = false? that shouldn't even execute, that is why it is inside the IF.

mcollina: I mean, just move `var autologging` at the top of the function.

katlimruiz: fixed

mcollina:
Thanks! Let's wait for @davidmarkclements or @jsumners if they are happy with the change!

test.js
```diff
@@ -292,6 +293,74 @@ test('no auto logging with autoLogging set to false', function (t) {
   }, handle)
 })
 
+test('no auto logging with autoLogging set to true and path ignored', function (t) {
+  var dest = split(JSON.parse)
+  var logger = pinoHttp({
+    autoLogging: true,
+    autoLoggingOptions: {
+      ignorePaths: ['^/ignorethis$']
```
jsumners: Should there be a test that includes a regex literal?

logger.js
```diff
@@ -61,13 +69,23 @@ function pinoLogger (opts, stream) {
...
     req.id = genReqId(req)
     req.log = res.log = logger.child({req: req})
     res[startTime] = res[startTime] || Date.now()
-    if (!req.res) { req.res = res }
 
     if (autoLogging) {
-      res.on('finish', onResFinished)
+      if (req.url && autoLoggingIgnorePaths.length) {
+        for (let i = 0; shouldLogSuccess && i < autoLoggingIgnorePaths.length; i++) {
+          shouldLogSuccess = req.url.match(autoLoggingIgnorePaths[i]) == null
```
jsumners: This could be potentially be a large number of regular expression invocations. I would theorize that this could drastically slow down request processing.

Would a simple `autoLoggingIgnorePaths.includes(i)` be better?

katlimruiz: Yes, I can do that change.
I thought of regex to give more flexibility to the developer, but I guess that it would be dumb for someone to put "/" as it'd be just better to turn off the autologging. So it is developer responsibility in the end.

The correct would be 
```
shouldLogSuccess = req.url.includes(autoLoggingIgnorePaths[i]) == null
```

katlimruiz: This is one solution (using default Node URL class):
```
      if (req.url && autoLoggingIgnorePaths.length) {
        var url = new URL(req.url);
        shouldLogSuccess = !autoLoggingIgnorePaths.includes(url.pathname);
      }
```

katlimruiz: This is the other
```
      if (req.url && autoLoggingIgnorePaths.length) {
        for (let i = 0; shouldLogSuccess && i < autoLoggingIgnorePaths.length; i++) {
          shouldLogSuccess = req.url.includes(autoLoggingIgnorePaths[i]) == null
        }
      }
```

katlimruiz: However, if you put regex (which I agree is much slower), then we could at some point ignore:
`/myproduct/123/update`

with the two new solutions, this would not be possible.

I think it depends. Perhaps if we stay with simple urls, then we can see how it goes. If someone asks for more complexity, then we can add it later on.

test.js
katlimruiz: this gets solved with next review (where regex are no longer used)

logger.js
katlimruiz: I went with the URL parsing solution. It solves the immediate need (health probes) and is fastest.

mcollina: Have you measured this? From my tests creating URL object was extremely slow.

katlimruiz: check https://github.com/Juntoz/test-url-parse
Indeed you were right. I replaced URL by fast-url-parser.

katlimruiz:
@davidmarkclements I would appreciate your comments to push this to master so this can be published to be updated on koa-pino-logger, and then apply to the application I'm building.
I know I could apply tarballs and so on, but is not as streamlined as wanted, plus I think is better to give to community so others can take advantage of the changes.

katlimruiz:
@davidmarkclements done, please let me know if ok.

logger.js
```diff
@@ -30,6 +30,7 @@ function pinoLogger (opts, stream) {
   delete opts.stream
 
   var autoLogging = (opts.autoLogging !== false)
+  var autoLoggingIgnorePaths = ((opts.autoLogging || {}).ignorePaths || [])
```
davidmarkclements:
```suggestion
  var autoLoggingIgnorePaths = (opts.autoLogging && opts.autoLogging.ignorePaths) ? opts.autoLogging  : []
```

katlimruiz: done

katlimruiz:
@davidmarkclements could you please review the last change? Hope you could, and thanks for your time and help.

katlimruiz:
hi @davidmarkclements could you help me giving the last review? Sorry to insist :embarrassed:

davidmarkclements:
yep thanks @katlimruiz LGTM!

PR #80, opened 2020-03-16
Add getPath option to autoLogging when req.url is not suitable for matching

arvinsingla:
Hello, thanks for the awesome module!

This PR adds a new callback function `getPath` to the autoLogging options which allows implementers to supply a path when checking against `ignorePaths` instead of relying on the default `req.url`.

This issue came up when we tried to implement pino-http into our Nestjs application using the [nestjs-pino](https://github.com/iamolegga/nestjs-pino) library and found that due to how the middleware is injected express was stripping the `req.url` as [documented here](https://expressjs.com/en/api.html#req.originalUrl). This led me to believe the best way to get the functionality we needed was to allow the implementer to supply the current path as a callback function.

An example of how we would use this in our config.
```
{
  autoLogging: {
    ignorePaths: ['/health'],
    getPath: req => req.originalUrl
 }
}
```

FYI I also put up this PR as I noticed the typescript definitions were out of date with the most recent release. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43137


arvinsingla:
@mcollina Just curious how many +1s are needed?

Issue #79, opened 2020-03-10
customLogLevel does not support 'silent'


sdavids:
https://github.com/pinojs/pino/blob/HEAD/docs/api.md#level-string
https://www.npmjs.com/package/express-pino-logger#api

```
'use strict';

const isProduction = process.env.NODE_ENV === 'production';

const pino = require('pino')();

const expressPinoConfig = {
  logger: pino,
  customLogLevel: (res, err) => {
    if (isProduction && res.statusCode === 404) {
      return 'silent';
    }
    return 'info';
  },
};

const expressPino = require('express-pino-logger')(expressPinoConfig);
```

Will lead to:

```
TypeError: log[level] is not a function
    at ServerResponse.onResFinished (/<redacted>/node_modules/pino-http/logger.js:58:15)
    at ServerResponse.emit (events.js:323:22)
    at onFinish (_http_outgoing.js:730:10)
    at onCorkedFinish (_stream_writable.js:721:5)
    at afterWrite (_stream_writable.js:528:5)
    at afterWriteTick (_stream_writable.js:515:10)
    at processTicksAndRejections (internal/process/task_queues.js:83:21)
```

Currently, there is no way to prevent requests having a specific status code from being logged.

mcollina:
Would you like to send a PR? I think we should just add a `pino.silent()` that's just an empty function.

baterson:
@mcollina Hello! I opened a PR which solves this issue https://github.com/pinojs/pino-http/pull/84 .
I hope I understood correctly how you wanted to solve it.

iandrc:
@mcollina I guess this one was fixed by https://github.com/pinojs/pino/pull/802

PR #84, opened 2020-03-22
Add support for "silent" log level

baterson:
Hello! PR fixes problem from this issue https://github.com/pinojs/pino-http/issues/79


mcollina:
This PR should be open against pino itself, not pino-http.

PR #85, opened 2020-03-22
Add support for custom success and error messages

baterson:
Hello! My team faced with an Issue recently. We wanted a detailed information on success and error log messages so it would be easier for us to filter it in datadog service.
Currently the messages are "request completed" and "request errored" and there is no way to configure them so I decided to write them as a callbacks which accepts response and error objects.
If you ok with those changes, I'll add tests and update README in this PR.


baterson:
Sure. I'll add tests and update README

baterson:
> Would you mind to add a unit test?
I added tests and updated the README

baterson:
> The API section of the readme has not been updated to reflect the new options.
I updated the README

Issue #87, opened 2020-03-31
Provide custom names for req/res log attributes.


CalumDickson:
Hi pino-http team,

This middleware is perfect for our product, however due to an agreed logging standard within our organisation, we need to rename `req` and `res` to `httpReq` and `httpRes` within our logs.

We've managed to hack around the issue by mutating the log message after the fact. However, it would be far cleaner if we could control these currently hard-coded attribute names with config.

I can raise a PR if this is something you would consider accepting?

mcollina:
Sure thing, send a PR

mcollina:
done in v4.5.0.

PR #88, opened 2020-03-31
Option to override the log object key naming with custom values

CalumDickson:
Adds a configuration option to rename the keys in the JSON log. Keys which can be renamed are `req`, `res`, `err`, and `responseTime`.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -62,31 +62,29 @@ function pinoLogger (opts, stream) {
     var log = this.log
     var responseTime = Date.now() - this[startTime]
     var level = customLogLevel ? customLogLevel(this, err) : useLevel
-    var payload = {}
-
-    payload[resKey] = this
-    payload[responseTimeKey] = responseTime
 
     if (err || this.err || this.statusCode >= 500) {
       var error = err || this.err || new Error('failed with status code ' + this.statusCode)
-      payload[errKey] = error
 
-      log[level](payload, errorMessage(error, this))
+      log[level]({
+        [resKey]: this,
+        [errKey]: error,
+        [responseTimeKey]: responseTime
+      }, errorMessage(error, this))
       return
     }
 
-    log[level](payload, successMessage(this))
+    log[level]({
+      [resKey]: this,
+      [responseTimeKey]: responseTime
+    }, successMessage(this))
   }
 
   function loggingMiddleware (req, res, next) {
     var shouldLogSuccess = true
 
     req.id = genReqId(req)
-
-    var childPayload = {}
-    childPayload[reqKey] = req
-
-    req.log = res.log = logger.child(childPayload)
+    req.log = res.log = logger.child({[reqKey]: req})
     res[startTime] = res[startTime] || Date.now()
 
     if (autoLogging) {
--- a/logger.js
+++ b/logger.js
@@ -12,10 +12,23 @@ function pinoLogger (opts, stream) {
   }
 
   opts = opts || {}
+
+  opts.customAttributeKeys = opts.customAttributeKeys || {}
+  var attributeKeys = {
+    req: opts.customAttributeKeys.req || 'req',
+    res: opts.customAttributeKeys.res || 'res',
+    err: opts.customAttributeKeys.err || 'err',
+    responseTime: opts.customAttributeKeys.responseTime || 'responseTime'
+  }
+  delete opts.customAttributeKeys
+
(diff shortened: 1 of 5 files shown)


CalumDickson:
Addresses issue #87

logger.js
```diff
@@ -12,10 +12,23 @@ function pinoLogger (opts, stream) {
...
   opts = opts || {}
+
+  opts.customAttributeKeys = opts.customAttributeKeys || {}
+  var attributeKeys = {
+    req: opts.customAttributeKeys.req || 'req',
+    res: opts.customAttributeKeys.res || 'res',
+    err: opts.customAttributeKeys.err || 'err',
+    responseTime: opts.customAttributeKeys.responseTime || 'responseTime'
+  }
+  delete opts.customAttributeKeys
```
mcollina, line 23: Can you avoid the delete ?

logger.js
```diff
@@ -12,10 +12,23 @@ function pinoLogger (opts, stream) {
...
 
   opts = opts || {}
+
+  opts.customAttributeKeys = opts.customAttributeKeys || {}
+  var attributeKeys = {
+    req: opts.customAttributeKeys.req || 'req',
+    res: opts.customAttributeKeys.res || 'res',
+    err: opts.customAttributeKeys.err || 'err',
+    responseTime: opts.customAttributeKeys.responseTime || 'responseTime'
+  }
```
mcollina, line 22: Would you mind to remove the wrapping object?

logger.js
```diff
@@ -51,29 +64,31 @@ function pinoLogger (opts, stream) {
     var log = this.log
     var responseTime = Date.now() - this[startTime]
     var level = customLogLevel ? customLogLevel(this, err) : useLevel
+    var payload = {}
```
mcollina, line 67: Please define all properties in one statement, as it will create a single shape.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pino-http",
-  "version": "4.4.0",
+  "version": "4.5.0",
```
mcollina, line 3: Please revert this, we will change when we release.

package.json
```diff
@@ -12,6 +12,7 @@
     "autocannon": "^2.4.1",
     "coveralls": "^3.0.0",
     "http-ndjson": "^3.1.0",
+    "pino-pretty": "^3.6.1",
```
mcollina, line 15: Is this needed?

logger.js
CalumDickson: I added this to be consistent the other pino-http specific options, which are all `delete`-ed after consumption. My assumption was that this is to prevent them being passed into the wrapped pino instance. I can of course remove this if not required?

logger.js
CalumDickson: Done, pending commit.

package.json
CalumDickson: When running `node example.js | pino` as per the README, I received the following:

`'pino' cli has been removed. Use 'pino-pretty' cli instead.`

I guess it's not _technically_ a devDependency, as they would need a global install to run the README example without npx prefix. For consistency I'll revert this.

package.json
CalumDickson: Done, pending commit

logger.js
CalumDickson:
```
var payload = {
  [resKey]: this,
  [responseTimeKey]: responseTime,
  [errKey]: error
}
```

As the attribute names are now dynamic, we'd need computed property syntax to do this in one statement, but this is only compatible with node > 4. Is that acceptable?

mcollina: yes it is.

logger.js
```diff
@@ -51,29 +62,31 @@ function pinoLogger (opts, stream) {
...
   }
 
   function loggingMiddleware (req, res, next) {
     var shouldLogSuccess = true
 
     req.id = genReqId(req)
-    req.log = res.log = logger.child({req: req})
+
+    var childPayload = {}
+    childPayload[reqKey] = req
```
mcollina, line 87: same

logger.js
CalumDickson: Great, done.

logger.js
CalumDickson: Done

logger.js
mcollina: Let's leave, not really required right now. This needs a bigger clean up.

PR #96, opened 2020-04-18
[#95] Add option to prevent serializers from being wrapped

### LINKED ISSUE #95: Ignores custom serializers on `opts.logger`
I have a global `logger` which is used throughout my application with a custom `err` serializer. 

This middleware [appears to always override it](https://github.com/pinojs/pino-http/blob/master/logger.js#L29) when passing in my logger which is quite unexpected, e.g.

```js
const app = require("express")();
const pino = require("pino");
const pinoMiddleware = require("express-pino-logger");

const logger = pino({
  prettyPrint: true,
  serializers: {
    err: (err) => {
      return {
        message: err.message,
      };
    },
  },
});

app.use(
  pinoMiddleware({
    logger,
  })
);

// This correctly uses my custom err serializer
logger.error({ err: new Error("b0rked 1") });

app.get("/", function (req, res) {
  // This does not use my custom err serializer?!
  req.log.error({ err: new Error("b0rked 2") });
});

app.listen(3000);
```

richardscarrott:
Fixes #95

Summary
1) Add `opts.wrapSerializers` to allow consumers to disable wrapping them with the standard serializers
   - This importantly ensures serializers applied to `opts.logger` are not always overridden with the default serializers.
2) Avoid mutating `opts` and `opts.serializers`

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,10 @@ var logger = require('pino-http')({
     res: pino.stdSerializers.res
   },
 
+  // Prevent serializers from being wrapped with std serializers
+  // (useful if you want to honor the serializers defined by `opts.logger`)
+  wrapSerializers: false,
+
   // Logger level is `info` by default
   useLevel: 'info',
 
@@ -259,7 +263,6 @@ by `foo`. In order to show these properties, along with the standard serialized
 properties, in the resulting logs, we can supply a serializer like:
 
 ```js
-var http = require('http')
 var logger = require('pino-http')({
   serializers: {
     req (req) {
@@ -274,10 +277,11 @@ var logger = require('pino-http')({
 })
 ```
 
-If you prefer to work with the raw value directly, you can pass in `opts.wrapSerializers` as `false`:
+If you prefer to work with the raw value directly, or you want to honor the custom
+serializers already defined by `opts.logger`, you can pass in `opts.wrapSerializers`
+as `false`:
 
 ```js
-var http = require('http')
 var logger = require('pino-http')({
   wrapSerializers: false,
   serializers: {
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ $ node example.js | pino-pretty
 * `customSuccessMessage`: set to a `function (res) => { /* returns message string */ }` This function will be invoked at each successful response, setting "msg" property to returned string. If not set, default value will be used.
 * `customErrorMessage`: set to a `function (res, err) => { /* returns message  string */ }` This function will be invoked at each failed response, setting "msg" property to returned string. If not set, default value will be used.
 * `customAttributeKeys`: allows the log object attributes added by `pino-http` to be given custom keys. Accepts an object of format `{ [original]: [override] }`. Attributes available for override are `req`, `res`, `err`, and `responseTime`. 
-* `wrapSerializers`: when false, custom serializers will be passed the raw value directly. Defaults to true
+* `wrapSerializers`: when `false`, custom serializers will be passed the raw value directly. Defaults to `true`.
 
 `stream`: the destination stream. Could be passed in as an option too.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -273,6 +274,23 @@ var logger = require('pino-http')({
...
+    req (req) {
+      // `req` is the raw `IncomingMessage` object, not the already serialized request from `pino.stdSerializers.req`.
+      return {
+        message: req.foo
+      };
+    }
+  }
+})
+```
+
```
richardscarrott, line 293: I wasn't quite sure of the best name for the option, some ideas were:

wrapSerializers: false
preSerialize: false
raw: true
serializeRaw: true

logger.js
```diff
@@ -20,13 +20,16 @@ function pinoLogger (opts, stream) {
...
-  opts.serializers = opts.serializers || {}
-  var requestSerializer = opts.serializers[reqKey] || opts.serializers.req || serializers.req
-  var responseSerializer = opts.serializers[resKey] || opts.serializers.res || serializers.res
-  var errorSerializer = opts.serializers[errKey] || opts.serializers.err || serializers.err
-  opts.serializers[reqKey] = serializers.wrapRequestSerializer(requestSerializer)
-  opts.serializers[resKey] = serializers.wrapResponseSerializer(responseSerializer)
-  opts.serializers[errKey] = serializers.wrapErrorSerializer(errorSerializer)
+  opts.wrapSerializers = 'wrapSerializers' in opts ? opts.wrapSerializers : true
+  if (opts.wrapSerializers) {
+    opts.serializers = Object.assign({}, opts.serializers) || {}
```
richardscarrott, line 25: I'm cloning the serializers here as mentioned in the pinojs#95

mcollina: It would actually be better to clone the full object as well, it should not delete options.

richardscarrott: Do you mean the full ‘opts’ object?

richardscarrott: I've additionally cloned the `opts` object, so the `delete` operators won't cause issues.

richardscarrott: @mcollina I just tested this in my application and noticed `wrapSerializers` is output on the log itself! I'm guessing I need to temporarily remove it in `wrapChild` in the same way `opts.logger` and `opts.genReqId` is done?

README.md
```diff
@@ -104,6 +104,7 @@ $ node example.js | pino-pretty
 * `customSuccessMessage`: set to a `function (res) => { /* returns message string */ }` This function will be invoked at each successful response, setting "msg" property to returned string. If not set, default value will be used.
 * `customErrorMessage`: set to a `function (res, err) => { /* returns message  string */ }` This function will be invoked at each failed response, setting "msg" property to returned string. If not set, default value will be used.
 * `customAttributeKeys`: allows the log object attributes added by `pino-http` to be given custom keys. Accepts an object of format `{ [original]: [override] }`. Attributes available for override are `req`, `res`, `err`, and `responseTime`. 
+* `wrapSerializers`: when false, custom serializers will be passed the raw value directly. Defaults to true
```
jsumners, line 107:
```suggestion
* `wrapSerializers`: when `false`, custom serializers will be passed the raw value directly. Defaults to `true`.
```

README.md
```diff
@@ -131,6 +132,10 @@ var logger = require('pino-http')({
     res: pino.stdSerializers.res
   },
 
+  // Prevent serializers from being wrapped with std serializers
+  // (useful if you want to honor the serializers defined by `opts.logger`)
+  wrapSerializers: false,
```
jsumners, line 137:
```suggestion
  // Set to `false` to prevent standard serializers from being wrapped.
  wrapSerializers: true,
```

logger.js
mcollina: Yes, exactly

README.md
mcollina: `wrapSerializers` is ok.

mcollina:
@jsumners are you ok?

PR #99, opened 2020-04-23
Remove trailing space in error message

LinusU:
Currently causing the tests to fail in express-pino-logger:

```text
  not ok message is set
    --- wanted                  
    +++ found                   
    -"request errored"          
    +"request errored "         
```


LinusU:
@mcollina done 👍

PR #100, opened 2020-04-28
Support regex path matching for autoLogging.ignorePaths option

jafern14:
I was using the latest version and have a use-case where I want to ignore a dynamic range of paths that can change over time. In order to support this I thought that adding RegExp support to the ignoring mechanism might be an option.

Let me know what you think. Thanks for taking a look.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -101,7 +101,11 @@ function pinoLogger (opts, stream) {
         }
         if (url && url.pathname) {
           shouldLogSuccess = !autoLoggingIgnorePaths.find(ignorePath => {
-            return new RegExp(ignorePath).test(url.pathname)
+            if (ignorePath instanceof RegExp) {
+              return ignorePath.test(url.pathname)
+            }
+
+            return ignorePath === url.pathname
           })
         }
       }
--- a/logger.js
+++ b/logger.js
@@ -100,7 +100,9 @@ function pinoLogger (opts, stream) {
           url = URL.parse(req.url)
         }
         if (url && url.pathname) {
-          shouldLogSuccess = !autoLoggingIgnorePaths.includes(url.pathname)
+          shouldLogSuccess = !autoLoggingIgnorePaths.find(ignorePath => {
+            return new RegExp(ignorePath).test(url.pathname)
+          })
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pino-http",
-  "version": "5.1.0",
+  "version": "5.2.0",
```
mcollina, line 3: Please rever this.

logger.js
```diff
@@ -100,7 +100,9 @@ function pinoLogger (opts, stream) {
           url = URL.parse(req.url)
         }
         if (url && url.pathname) {
-          shouldLogSuccess = !autoLoggingIgnorePaths.includes(url.pathname)
+          shouldLogSuccess = !autoLoggingIgnorePaths.find(ignorePath => {
+            return new RegExp(ignorePath).test(url.pathname)
+          })
```
mcollina, line 105: We should not be generating a new regexp for strings that are anot already regexp.

jafern14: @mcollina let me know if you prefer some other approach. I thought this would work as well and be a bit more explicit.

mcollina:
@jsumners @davidmarkclements wdyt?

logger.js
jsumners: This is taking user input (`url.pathname`) and creating a new regular expression from it. See https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS

mcollina: This has been already fixed.

katlimruiz: 👍

jsumners:
ping: @davidmarkclements

DanHulton:
Is there any intention of merging this?  I can always target a branch to get this functionality, but it would be a lot cleaner to just be able to use the base package.

I'm not clear on the governance of this project, does @davidmarkclements need to +1 it, because it's been many months waiting on him to check in.

davidmarkclements (approved):
lgtm

davidmarkclements:
published as 5.3.0 sorry for the delay everyone

PR #104, opened 2020-06-07
Fix incorrect documentation for customErrorMessage 

Tadimsky:
The Readme currently defines the `customErrorMessage` function as having the following signature:

```js
function (res, err) => { /* returns message  string */ }
```

However, the actual implementation
https://github.com/pinojs/pino-http/blob/d593aa6084f1fe70a506ad8f202fc7f339b87fb5/logger.js#L79

Uses `err, res`.

This changes fixes the Readme so that people can use it the right way.
I also added the error to the message in the test to show the use cases for it

I also removed the `package-lock.json` file from the `.gitignore` as that should be checked into source control. Please let me know if it's removed by design.

There is also an error in the `@types/pino-http` package added in https://github.com/DefinitelyTyped/DefinitelyTyped/commit/770b1753b685c22a2a75309ed45058cd66159adb
I'll be fixing that as well.

### DIFF
--- a/test.js
+++ b/test.js
@@ -671,7 +671,7 @@ test('uses the custom errorMessage callback if passed in as an option', function
   var customErrorMessage = 'Custom error message'
   var logger = pinoHttp({
     customErrorMessage: function (err, res) {
-      return customErrorMessage
+      return customErrorMessage + ' ' + err.toString()
     }
   }, dest)
 
@@ -681,7 +681,7 @@ test('uses the custom errorMessage callback if passed in as an option', function
   })
 
   dest.on('data', function (line) {
-    t.equal(line.msg, customErrorMessage)
+    t.contains(line.msg, customErrorMessage)
     t.end()
   })
 })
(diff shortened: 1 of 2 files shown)


Tadimsky:
Fixed the `@types/pino-http` package as well: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45344

.gitignore
```diff
@@ -39,7 +39,4 @@ profile*
 # nyc
 .nyc_output
 
-.DS_Store
-
-# lockfile
-package-lock.json
```
mcollina, line 45: Please restore this and delete package-lock.json.

test.js
```diff
@@ -681,7 +681,8 @@ test('uses the custom errorMessage callback if passed in as an option', function
   })
 
   dest.on('data', function (line) {
-    t.equal(line.msg, customErrorMessage)
+    console.log(line.msg)
```
mcollina, line 684: remove this

Tadimsky: whoops, my bad, was just testing something when doing this. thanks, will remove

.gitignore
mcollina: not on modules

PR #113, opened 2020-09-30
Log errors for ignored paths

pguelpa:


logger.js
```diff
@@ -80,15 +81,16 @@ function pinoLogger (opts, stream) {
       return
     }
 
-    log[level]({
-      [resKey]: this,
-      [responseTimeKey]: responseTime
-    }, successMessage(this))
+    if (this.shouldLogSuccess || (autoLoggingStatusCodeThreshold && this.statusCode >= autoLoggingStatusCodeThreshold)) {
```
mcollina, line 84: I don't think `this.shouldLogSuccess` is actually needed, can you please move it back?

logger.js
```diff
@@ -49,6 +49,7 @@ function pinoLogger (opts, stream) {
   var autoLogging = (opts.autoLogging !== false)
   var autoLoggingIgnorePaths = (opts.autoLogging && opts.autoLogging.ignorePaths) ? opts.autoLogging.ignorePaths : []
   var autoLoggingGetPath = opts.autoLogging && opts.autoLogging.getPath ? opts.autoLogging.getPath : null
+  var autoLoggingStatusCodeThreshold = opts.autoLogging && opts.autoLogging.statusCodeThreshold ? opts.autoLogging.statusCodeThreshold : null
```
mcollina, line 52:
```suggestion
  var autoLoggingStatusCodeThreshold = opts.autoLogging && opts.autoLogging.statusCodeThreshold ? opts.autoLogging.statusCodeThreshold : 0
```

logger.js
pguelpa: The main reason to have it run on every finish is to capture the error handling logic on line 72, without it errors on ignored paths aren't logged unless the request receives an `error` event (and from the tests it looks like the client must trigger that manually?).

The two new tests `auto logging with autoLogging set to true and path ignored and error` and `auto logging with autoLogging set to true and path ignored and custom status code threshold` try to capture this and will fail with the old logic.

mcollina:
@jsumners wdyt of this?

README.md
```diff
@@ -100,6 +100,7 @@ $ node example.js | pino-pretty
 * `autoLogging`: set to `false`, to disable the automatic "request completed" and "request errored" logging. Defaults to `true`. If set to an object, you can provide more options.
 * `autoLogging.ignorePaths`: array that holds one or many paths that should not autolog on completion. Paths will be matched exactly to the url path `req.url` (using Node class `URL.pathname`). This is useful for ignoring e.g. health check paths that get called every X seconds, and would fill out the logs unnecessarily. If the path matches and succeeds (http 200), it will not log any text. If it fails, it will log the error (as with any other path).
 * `autoLogging.getPath`: set to a `function (req) => { /* returns path string */ }`. This function will be invoked to return the current path as a string. This is useful for checking `autoLogging.ignorePaths` against a path other than the default `req.url`. e.g. An express server where `req.originalUrl` is preferred.
+* `autoLogging.statusCodeThreshold`: set a status code threshold above which responses that would normally be ignored will be logged.
```
jsumners, line 103: There should be a note describing the default threshold.

PR #114, opened 2020-10-09
Adds missing 4xx errors to custom message handlers

squareloop1:
Currently only errors with status code >= 500 and success messages without the error object can be handled by `customSuccessMessage` and `customErrorMessage`.

This adds an optional error param to `customSuccessMessage` which contains http errors (4xx) or `undefined` if no error exists.

The reason for this is to be able to log details about the http exceptions for investigation. With the current implementation it is only possible to modify the message and log details about 5xx errors.

An alternative to this would be to handle 4xx errors in `customErrorMessage` by setting `this.statusCode >= 400` in this conditional:  https://github.com/pinojs/pino-http/blob/4d1c32038c8d287befe41f8be096047b77b1435f/logger.js#L72
But this would be a breaking change, so I guess its better to handle 4xx errors in `customSuccessMessage` like it would be possible with this PR.


mcollina (approved):
lgtm

test.js
```diff
@@ -674,7 +674,7 @@ test('uses the custom successMessage callback if passed in as an option', functi
   var dest = split(JSON.parse)
   var customResponseMessage = 'Custom response message'
   var logger = pinoHttp({
-    customSuccessMessage: function (res) {
+    customSuccessMessage: function (res, err) {
```
jsumners, line 677: Where is the test for the `err` possible values?

squareloop1: @jsumners Added a test for the optional param

test.js
```diff
@@ -690,6 +690,27 @@ test('uses the custom successMessage callback if passed in as an option', functi
   })
 })
 
+test('custom successMessage has optional error parameter', function (t) {
+  var dest = split(JSON.parse)
+  var customErrorMessage = 'Custom response message'
+  var logger = pinoHttp({
+    customSuccessMessage: function (res, err) {
+      var optionalErrorMessage = err ? err.toString() : 'error is undefined'
+      return customErrorMessage + ' ' + optionalErrorMessage
```
jsumners, line 699: We should verify that the `err` object is present. Otherwise the test is invalid:

```suggestion
      t.ok(err)
      var optionalErrorMessage = err.toString()
      return customErrorMessage + ' ' + optionalErrorMessage
```

test.js
```diff
@@ -690,6 +690,27 @@ test('uses the custom successMessage callback if passed in as an option', functi
...
+    }
+  }, dest)
+
+  setup(t, logger, function (err, server) {
+    t.error(err)
+    doGet(server, '/404')
+  })
+
+  dest.on('data', function (line) {
+    t.contains(line.msg, customErrorMessage)
```
jsumners, line 709:
```suggestion
    t.is(line.msg, `${customErrorMessage} <the text from the error message>`)
```

squareloop1: Yes you're right, this test would not make much sense without testing for an actual error. I used the logic from the original implementation of `customErrorMessage` and it seems like this test is not making any sense either as its only checking for the initial message too:
https://github.com/pinojs/pino-http/blob/4d1c32038c8d287befe41f8be096047b77b1435f/test.js#L693-L711

I tried adding a 4xx error but I'm not quite sure how to achieve it as something like this:
```JS
  function handle (req, res) {
    logger(req, res)
    res.emit('error', new Error('Some error'))
    res.end()
  }

  setup(t, logger, function (err, server) {
    t.error(err)
    doGet(server, '/404')
  }, handle)
```
would just run into the default error handling and respond with a `request errored`.

The test is making me question if this PR is really making any sense as I'm not quite sure if the 4xx errors I wanted to handle initially with this implementation are more of an issue to solve in the application framework where I am consuming pino-http.

jsumners: So what is your conclusion?

squareloop1:
I'm going to close this PR for now as I'm not quite sure of the benefits for this as the errors I initially thought to catch with it are handled better in the upstream framework anyways.

Issue #116, opened 2020-11-04
Missing log on client timeout


dalexanco:
I have an issue : when my client cut the connection (generally because of a timeout), then pino-http does not log anything. 

The default behavior of express is to continue the request processing, even if the client has gone, so it could be interesting to have these logs. 

A small example :
```javascript
const express = require('express')
const pinoLogger = require('pino-http')
const pino = require('pino')

const app = express()
const pinoInstance = pino({
  prettyPrint: true,
})
app.use(pinoLogger({ logger: pinoInstance }))

app.get('/foo', (req, res) => {
  pinoInstance.info('Received request')
  setTimeout(() => {
    res.sendStatus(200)
    pinoInstance.info('Response sent')
  }, 10000)
})

app.use((err, req, res, next) => {
  pinoInstance.error('Error : ', err)
})

app.listen(3000, () => {
  pinoInstance.info('Server ready')
})
```

**Case 1 : a simple request, no timeout**
```bash
[1604486775436] INFO     (15596 on CM000091.local): Server ready
[1604486778910] INFO     (15596 on CM000091.local): Received request
[1604486788921] INFO     (15596 on CM000091.local): Response sent
[1604486788922] INFO     (15596 on CM000091.local): request completed
    res: {
      "statusCode": 200,
      "headers": {
        "x-powered-by": "Express",
        "content-type": "text/plain; charset=utf-8",
        "content-length": "2",
        "etag": "W/\"2-nOO9QiTIwXgNtWtBJezz8kv3SLc\""
      }
    }
    responseTime: 10012
    req: {
      "id": 1,
      "method": "GET",
      "url": "/foo",
      "headers": {
        "accept": "application/json, text/plain, */*",
        "user-agent": "axios/0.21.0",
        "host": "localhost:3000",
        "connection": "close"
      },
      "remoteAddress": "::ffff:127.0.0.1",
      "remotePort": 59200
    }
`
``

**Case 2 : a request with a timeout set (3sec)**
```
[1604487109825] INFO     (15694 on CM000091.local): Server ready
[1604487111452] INFO     (15694 on CM000091.local): Received request
[1604487121458] INFO     (15694 on CM000091.local): Response sent
```
No pino-http log "request completed". 

Is it a normal behavior ?

PR #117, opened 2020-11-04
feat: log even on closed connections

dalexanco:
Hi, this PR change the event used by pino-http on response from 'finish' to 'close'.
It let the opportunity to log request which does not have open socket anymore. 

I add a new option "onlyLogFinishedRequest" set by default to true (to keep existing behavior).

More information : https://github.com/pinojs/pino-http/issues/116

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -65,15 +65,20 @@ function pinoLogger (opts, stream) {
 
   function onResFinished (err) {
     this.removeListener('error', onResFinished)
+<<<<<<< Updated upstream
+=======
+    this.removeListener('finish', onResFinished)
+>>>>>>> Stashed changes
     this.removeListener('close', onResFinished)
 
     var log = this.log
+    var requestLogged = !!this.requestLogged
     var responseTime = Date.now() - this[startTime]
     var level = customLogLevel ? customLogLevel(this, err) : useLevel
 
     if (err || this.err || this.statusCode >= 500) {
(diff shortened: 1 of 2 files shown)


logger.js
```diff
@@ -63,15 +65,20 @@ function pinoLogger (opts, stream) {
 
   function onResFinished (err) {
     this.removeListener('error', onResFinished)
+<<<<<<< Updated upstream
+=======
     this.removeListener('finish', onResFinished)
+>>>>>>> Stashed changes
```
mcollina, line 71: It seems you have made some git mistakes :). Would you mind to fix them?

mcollina (changes requested):

logger.js
jmealo: @dalexanco: ^^

mcollina: @dalexanco feel free to send a fresh PR if you'd like.

jmealo: @mcollina: https://github.com/pinojs/pino-http/pull/123
@dalexanco: I really wanted to get this merged upstream, so, I checked out your branch and committed the fix with you as the co-author; all your other commits are left intact. Thanks for the fix!

dalexanco: my bad, I have been away for a while. 
Thank you for the 2nd PR !
Really smart to allow the feature only on v12, that was the real reason of my absence on this PR : I fail to make it work on v10 😅

PR #121, opened 2021-01-29
fix: add res to reqCustomProp function signature

### LINKED ISSUE #119: Add res properties to custom props
The function `reqCustomProps` only has access to the `req` object, but best practice is to store data scoped to the request in `res.locals`, to which there is no access in `reqCustomProps`.
Does `res.locals` automatically get logged?
If not I'd be happy to open a pull request as it seems like by either just adding the `res` parameter to the function, or adding another similar function named appropriately, it should have minimal impact on current users' implementations.

josh-hemphill:
Added `res` object as additional function parameter to the `reqCustomProps` function
Non-breaking now, but function name will need to be changed in next semver major update, will add PR after this is merged to reuse the branch.

Updated `README.md` and `test.js` with new function signature.

Side effects:

- had to add "bash" to beginning of benchmark command since it relies on bash, and errors out on systems where bash is not the default.
- added "fix" npm script since locally installed `standard` has to be run inside npm script to use local version, and `standard --fix` must be run to get tests to run.

closes #119


mcollina (approved):
lgtm

jsumners (approved):

davidmarkclements (approved):

davidmarkclements:
5.4.0 published

PR #122, opened 2021-01-31
either reqCustomProps or customProps are valid

josh-hemphill:
Since the function's name no longer accurately represents its new signature, whenever there are other breaking changes to be made to the interface, this would be a good semantic change to also include.


mcollina (changes requested):
Can you keep the previous name as an alias? In this way this will be a minor change.

mcollina (changes requested):
Can you please add a test to verify the old prop?

mcollina (approved):
lgtm

PR #123, opened 2021-02-20
Feat/log on closed connections

jmealo:
This PR fixes the git oops in #117 authored by @dalexanco 

Minor change:
I decided not to attach/detach a listener for the `finish` event when we don't plan on logging it. This shouldn't leave any dangling handlers or change the intended behavior.

Tests pass.

### DIFF
--- a/test.js
+++ b/test.js
@@ -7,6 +7,7 @@ var pino = require('pino')
 var split = require('split2')
 
 var ERROR_URL = '/make-error'
+var WAIT_URL = '/wait'
 var noop = function () {}
 
 function setup (t, logger, cb, handler) {
@@ -14,6 +15,9 @@ function setup (t, logger, cb, handler) {
     logger(req, res)
     if (req.url === '/') {
       res.end('hello world')
+      return
+    } else if (req.url === WAIT_URL) {
+      setTimeout(() => res.end('that was boring'), 300)
       return
     } else if (req.url === ERROR_URL) {
       res.statusCode = 500
@@ -53,9 +57,11 @@ test('default settings', function (t) {
   dest.on('data', function (line) {
     t.ok(line.req, 'req is defined')
     t.ok(line.res, 'res is defined')
+    t.ok(line.finished, 'finish status is defined')
     t.equal(line.msg, 'request completed', 'message is set')
     t.equal(line.req.method, 'GET', 'method is get')
     t.equal(line.res.statusCode, 200, 'statusCode is 200')
+    t.equal(line.finished, true, 'finished is true')
     t.end()
   })
 })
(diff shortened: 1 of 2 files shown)


mcollina:
I don't think writableEnded is there in node v10.

jmealo:
@mcollina: I verified this fails on Node 10 locally and passed up to 15. I think that you're on to something.

jmealo:
@mcollina: I reviewed (and you reviewed apparently :D) the PR where `writableEnded` landed: https://github.com/nodejs/node/pull/28934/files

I wasn't sure if we exited early based on the order of events firing, so I removed my listener tweak before I got started.

I tried to backport this to Node.js 10 using `this._writableState.ending` but it doesn't look like that's exposed to userland? Any thoughts?

jmealo:
@mcollina: I think this works best. Let me know what you think.

jmealo:
@mcollina: Here's what I landed on:

**Behavior:**
If a user tries to log unfinished connections and `writableEnded` isn't available; we throw:`'onlyLogFinishedRequest': false requires Node v12.9.0 or later` at start-up.

**Reasoning:**
Node 10 is EoL in April and it wasn't immediately obvious for me how to back-port this. We already throw on other invalid parameter combinations. This seems like reasonable behavior given the circumstances.

test.js
```diff
@@ -283,6 +289,64 @@ test('responseTime for request emitting error event', function (t) {
...
+    setup(t, logger, function (err, server) {
+      t.error(err)
+      var request = doGet(server, WAIT_URL)
+      request.on('error', function (error) {
+        t.equal(error.message, 'socket hang up')
+        setTimeout(() => {
+          var line = dest.read()
+          t.equal(line, null)
+          t.end()
+        }, 100)
```
mcollina, line 322: Can you use a `setImmediate()` instead?

test.js
```diff
@@ -283,6 +289,64 @@ test('responseTime for request emitting error event', function (t) {
...
+    setup(t, logger, function (err, server) {
+      t.error(err)
+      var request = doGet(server, WAIT_URL)
+      request.on('error', function (error) {
+        t.equal(error.message, 'socket hang up')
+        setTimeout(() => {
+          var line = dest.read()
+          t.equal(line.finished, false, 'unfinished request is logged and flagged')
+          t.end()
+        }, 100)
```
mcollina, line 343: Can you use a `setImmediate()` instead?

test.js
```diff
@@ -15,6 +16,9 @@ function setup (t, logger, cb, handler) {
     if (req.url === '/') {
       res.end('hello world')
       return
+    } else if (req.url === WAIT_URL) {
+      setTimeout(() => res.end('that was boring'), 300)
```
mcollina, line 20: why a setTimeout? Can you use a `setImmediate()` instead?

mcollina (changes requested):

test.js
jmealo: Hm... The tests were the only thing I didn't rewrite; I can take a look at these later today. I don't see why not.

jmealo: @mcollina: I wasn't able to completely replace `setTimeout` with `setImmediate` without a triple nested `setImmediate` which seemed worse from a clarity standpoint. The author's original intent was to give the request enough time to start so a failure could be simulated and ultimately logged. One or two of these `setTimeout` can be easily replaced with `setImmediate`. To completely remove `setTimeout` would require more work than I can set aside right now. I may be able to revisit later -- or anyone else is free to pick this up as well! It's a good task for those looking to learn about the particulars of the event loop.

mcollina: using setTimeout will lead to flaky CI runs, I really need this fixed before landing.

Another approach is using some logical events instead of waiting.

PR #126, opened 2021-04-15
Fix README documentation typo for stream

meliasson:
A missing newline made the stream documentation part of the last
bullet in the above bullet list.

Before fix:
before

After fix:
after


mcollina (approved):
LGTM

PR #156, opened 2021-09-16
feature: quietReqLogger option for a quieter child logger available on the request object

darrenmce:
Hi, I've recently started using this for our logging as we migrated from bunyan to pino.

A behaviour we used to have in the bunyan-land was that our request logging middleware would similarly provide a request-context aware child logger (available at `req.log`) however it would only contain the request id that you could use to filter logs on and see logs for the entire request transaction (including the one message with the full request serialized).

Instead of just requesting this feature, I took a stab at a PR instead - please let me know if this is something you'd consider.

Some things that I'm on the fence/unsure about:

1) the name of the option `quietReqLogger` :man_shrugging: 
2) the top-level request id being only populated when the quietReqLogger is enabled with no granular control
3) `res.log` and `req.log` are diverging with this option enabled, is there any downstream issues with this?
4) typescript definitions - not sure who manages these


mcollina (approved):
lgtm

jsumners (approved):

PR #168, opened 2021-10-22
fixed customProps to capture response object

vavar:
I try to use `customProps` to return `res.statusCode` in top-level with `nestjs-pino` but found statusCode is incorrect.
so I add below testcase to ensure not related to nest dependencies.

```
test('uses custom request properties to log additional attributes when response provided', function (t) {
  var dest = split(JSON.parse)
  function customPropsHandler (req, res) {
    if (req && res) {
      return {
        key1: 'value1',
        key2: res.statusCode
      }
    }
  }
  var logger = pinoHttp({
    reqCustomProps: customPropsHandler
  }, dest)

  setup(t, logger, function (err, server) {
    t.error(err)
    doGet(server, ERROR_URL)
  })

    t.equal(line.key1, 'value1')
    t.equal(line.key2, 500)
    t.end()
  })
})
```
so this fixed is correct `response` object to be pass through `customProps` option.

NOTE: not sure `Destructuring assignment` is available on this repo or not, please advise.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -74,7 +74,11 @@ function pinoLogger (opts, stream) {
     var responseTime = Date.now() - this[startTime]
     var level = customLogLevel ? customLogLevel(this, err) : useLevel
 
-    log = (typeof customProps === 'function') ? log.child(customProps(this[reqObject], this)) : log.child(customProps)
+    var req = this.log[reqObject]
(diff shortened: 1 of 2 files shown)


mcollina:
CI is failing

PR #177, opened 2021-11-17
Add request received event (customReceivedMessage)

nicolaichuk:
Change:

* add request received event (customReceivedMessage)
* add yarn.lock to .gitignore
* add t.end() to some test

Please review and merge to upstream.


jsumners (approved):
Looks good to me.

mcollina (approved):
lgtm

logger.js
```diff
@@ -133,6 +134,9 @@ function pinoLogger (opts, stream) {
       }
 
       if (shouldLogSuccess) {
+        const level = customLogLevel ? customLogLevel(res) : useLevel
```
mcollina, line 137: Could you add another check if `level` is undefined? In this case it should not log the next line.

logger.js
```diff
@@ -54,6 +54,7 @@ function pinoLogger (opts, stream) {
   const autoLoggingGetPath = opts.autoLogging && opts.autoLogging.getPath ? opts.autoLogging.getPath : null
   delete opts.autoLogging
 
+  const receivedMessage = opts.customReceivedMessage || function () { return 'request received' }
```
mcollina, line 57: I would prefer to have pino-http not log this message by default.

mcollina (changes requested):

logger.js
nicolaichuk: Fixed

logger.js
nicolaichuk: Fixed

nicolaichuk:
@mcollina, please review change.

mcollina (approved):
lgtm

Issue #175, opened 2021-11-03
Pino 7 TypeScript


rstrand:
Now pino 7 is including TypeScript types, would it make sense to include TypeScript types in this package as well? The types in DefinitelyTyped for this package don't work with the pino 7 types.

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

cc @kibertoad

rstrand:
I don't have any experience adding types to a project. Would it be acceptable to copy the types from definitely typed and fix the errors?

gkampitakis:
Hello 👋 are there any updates on this issue? Are types going to be included in `pino-http`? I wouldn't mind trying and start a pr with your help

mcollina:
A PR would be nice.

jarcodallo:
> A PR would be nice.

Hi @mcollina, I opened a PR with the type definition. let me know what do you think. 🚀

oddeirik:
Am I missing some required changes after the typings now being included in the package itself?

The IncomingMessage declaration for a 'req' property is missing in types now, which means TS will complain about doing `req.log.info()' in Express handlers/middleware.

This module declaration was included in the typings from @types/pino-http, but does not exist in the typings included in pino-http itself now:

```
declare module 'http' {
    interface IncomingMessage {
        id: PinoHttp.ReqId;
        log: Logger;
    }

    interface ServerResponse {
        err?: Error | undefined;
    }

    interface OutgoingMessage {
        [PinoHttp.startTime]: number;
    }
}
```

kibertoad:
@oddeirik Would you like to send a PR to fix this?

umarov:
https://github.com/pinojs/pino-http/pull/184 I think this one "removed" some PinoHttp.Options from types. They aren't being merged properly. I am using NestJS Pino Logger and using the redact option. It was yelling at me once I upgraded to the latest version. Moving back to `6.3.0` fixed it. I haven't figured out how that PR could have broken the types. I need to take a look again and see if I can make a PR.

jarcodallo:
> #184 I think this one "removed" some PinoHttp.Options from types. They aren't being merged properly. I am using NestJS Pino Logger and using the `redact` option.
> 
> ```ts
> LoggerModule.forRoot({
>       pinoHttp: [
>         {
>           redact: [
>             'req.headers.authorization',
>             'req.headers.cookie',
>             'req.headers["x-csrf-token"]'
>           ]
>         }
[...]

PR #178, opened 2021-11-18
Pino 7 Type Definition

jarcodallo:
Fix for #175

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,65 @@
+import { IncomingMessage, ServerResponse } from 'http';
+import pino from 'pino';
+
+export = PinoHttp;
+
+declare function PinoHttp(opts?: PinoHttp.Options, stream?: pino.DestinationStream): PinoHttp.HttpLogger;
+declare function PinoHttp(stream?: pino.DestinationStream): PinoHttp.HttpLogger;
+
+declare namespace PinoHttp {
+    interface HttpLogger {
+        (req: IncomingMessage, res: ServerResponse, next?: () => void): void;
+        logger: pino.Logger;
+    }
+    type ReqId = number | string | object;
+
+    interface Options extends pino.LoggerOptions {
+        logger?: pino.Logger | undefined;
+        genReqId?: GenReqId | undefined;
+        useLevel?: pino.Level | undefined;
+        stream?: pino.DestinationStream | undefined;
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

index.d.ts
```diff
@@ -0,0 +1,65 @@
+import { IncomingMessage, ServerResponse } from 'http';
+import pino from 'pino';
+
+export = PinoHttp;
+
+declare function PinoHttp(opts?: PinoHttp.Options, stream?: pino.DestinationStream): PinoHttp.HttpLogger;
+declare function PinoHttp(stream?: pino.DestinationStream): PinoHttp.HttpLogger;
+
+declare namespace PinoHttp {
```
kibertoad, line 9: AFAIR, namespaces are no longer a recommended TS pattern, using module is preferable.

index.d.ts
```diff
@@ -0,0 +1,65 @@
+import { IncomingMessage, ServerResponse } from 'http';
+import pino from 'pino';
+
+export = PinoHttp;
```
kibertoad, line 4: Can you do named exports instead? This style of export has issues when esModuleInterop is disabled, AFAIR.

index.d.ts
```diff
@@ -0,0 +1,65 @@
+import { IncomingMessage, ServerResponse } from 'http';
```
kibertoad, line 1: looks like you need to add node ts types as dev dependency

rstrand:
These types look copied from definitely typed (and updated/fixed) but with the attribution removed from the top of index.d.ts
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pino-http/index.d.ts

kibertoad:
@jarcodallo Could you please restore the attribution?

jarcodallo:
> These types look copied from definitely typed (and updated/fixed) but with the attribution removed from the top of index.d.ts https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pino-http/index.d.ts

Yeah I copy the definition from there and add mirror changes. I will restore the attribution.

jarcodallo:
> @jarcodallo Could you please restore the attribution?

I will, sorry for that. I just remove all the comments

kibertoad (approved):

PR #181, opened 2021-11-23
Export Options from index.d.ts

adarnon:
Exporting the Options interface makes writing TS code that uses pino-http much easier. Especially because the "first parameter" type can't be directly inferred to a TS `type` due to overloads.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -23,7 +23,7 @@ interface HttpLogger {
 }
 type ReqId = number | string | object;
 
-interface Options extends pino.LoggerOptions {
+export interface Options extends pino.LoggerOptions {
     logger?: pino.Logger | undefined;
     genReqId?: GenReqId | undefined;
     useLevel?: pino.Level | undefined;


mcollina (commented):
Could you add a test for these types? We use tsd.

hoonoh:
I was on my way to creating a similar PR to this. The problem I am having with the new type definition is with missing `HttpLogger` definition. 
[@types/pino-http](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pino-http/index.d.ts) used to have definitions using namespace which exposed all definitions within the namespace (`HttpLogger`, `ReqId`, `Options`, `GenReqId`, `AutoLoggingOptions`, `CustomAttributeKeys`, `StdSerializers`), and I think it would be better to either export all definitions as in `@types/pino-http` or reintroduce type definition with namespace.

jarcodallo:
> I was on my way to creating a similar PR to this. The problem I am having with the new type definition is with missing `HttpLogger` definition. [@types/pino-http](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pino-http/index.d.ts) used to have definitions using namespace which exposed all definitions within the namespace (`HttpLogger`, `ReqId`, `Options`, `GenReqId`, `AutoLoggingOptions`, `CustomAttributeKeys`, `StdSerializers`), and I think it would be better to either export all definitions as in `@types/pino-http` or reintroduce type definition with namespace.

My bad I didn't expose the interfaces, but I agree with @hoonoh we should expose the interfaces. @adarnon can you add the export for the other interfaces?

adarnon:
I can do it but will only have time for it next week probably, so if someone wants it done sooner you can get moving with it :)

hoonoh:
I added new PR #184 with tests ;)

adarnon:
@hoonoh Thank you :)

Issue #73, opened 2019-11-21
Add option for custom logging format


taylorjdawson:
Using the serializers I can customize what get's logged. However, I do not have control over the logging format. For example, I would like to log like so:
```
[1574212772844] INFO (35 on db805887c09f): 11.11.111.111 - GET /url/path - Status Code: 200
```
Essentially it would mean removing the object with `res` & `err` [here](https://github.com/pinojs/pino-http/blob/a7c82bc2d82847ba6692b20de1e7d2b65c4a1624/logger.js#L49-L60) in `logger.js`:
```
      log[level]({
        res: this,
        err: err || this.err || new Error('failed with status code ' + this.statusCode),
        responseTime: responseTime
      }, 'request errored')
      return
    }


    log[level]({
      res: this,
      responseTime: responseTime
    }, 'request completed')
```

...allowing the user to decide how to format the log.

mcollina:
I'm -1 in removing the lines from pino-http for _formatting_ reasons. I'm happy to review a PR that make things more flexible here.

taylorjdawson:
Okay I did some digging and I think what I am looking for is something like [this](https://github.com/pinojs/pino-http-print/blob/master/index.js). I think it would be possible by using the stream like you did to manipulate the logging format. Is that true? If so I guess it wouldn't apply to the `pino-http` module.

waynebloss:
I'd like this option as well.

The suggested `11.11.111.111 - GET /url/path - Status Code: 200` format would be preferred in any configuration instead of the default `request completed` message, however i'd also add an option to remove details of `req` and `res` unless I opt into getting them for specific endpoints.

This code looks easy enough to fork though, so I won't wait around too long for it.

waynebloss:
I couldn't get `pino-http-print` to work at all.

Attempt 1:

```ts
import express from "express";
import pino from "pino";
import expressPino from "express-pino-logger";
import pinoHttpPrint from "pino-http-print";

const log = pino({
  prettyPrint: {
    translateTime: "SYS:yyyy-mm-dd HH:MM:ss.l",
  },
});

const app = express();

app.use(expressPino(pinoHttpPrint()));
```

Attempt 2:

```ts
app.use(expressPino(pinoHttpPrint({ all: true })));
```

Attempt 3:

```ts
app.use(
  expressPino({
    logger: log,
    stream: pinoHttpPrint({ all: true }),
  }),
);
```

[...]

PR #182, opened 2021-11-24
add format example test

Eomm:
fixes https://github.com/pinojs/pino-http/issues/73

It is possible to get the desired output using the new pino 7 transport.
I have added an example with tests

### DIFF
--- a/example-custom-format.js
+++ b/example-custom-format.js
@@ -1,0 +1,24 @@
+'use strict'
+
+const http = require('http')
+const server = http.createServer(handle)
+
+const logger = require('./')({
+  quietReqLogger: true,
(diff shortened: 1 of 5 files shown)


example-custom-format.js
```diff
@@ -0,0 +1,24 @@
...
+    }
+  }
+})
+
+function handle (req, res) {
+  logger(req, res)
+  res.end('hello world')
+}
+
+server.listen(3000)
```
mcollina, line 24: It's better if you use standard error to print something out after the listen is complete and then use that as a trigger to send the incoming http message.

mcollina (approved):
lgtm

PR #186, opened 2021-11-27
Add back pino-related typings on IncomingMessage, OutgoingMessage and…

oddeirik:
… ServerResponse.

Also fixed a test that fails on Windows (no /dev/null available, will print to STDOUT instead).

Ref. #175.
These additional typings are included in the @types/pino-http, so I just added them back in.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,3 @@
-// Type definitions for pino-http 6.x
 // Project: https://github.com/pinojs/pino-http#readme
 // Definitions by: Christian Rackerseder <https://github.com/screendriver>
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for pino-http 6.0
+// Type definitions for pino-http 6.x
```
mcollina, line 1: Let's remove this line

mcollina (commented):
@kibertoad can you review?

kibertoad (approved):

mcollina (approved):
lgtm

PR #187, opened 2021-11-29
Add missing 'customReceivedMessage' property to options type definition

oddeirik:
Add missing property in type definition.


mcollina:
cc @kibertoad

kibertoad (approved):

PR #191, opened 2021-12-08
Early return on silent level

### LINKED ISSUE #190: How to ignore autoLogging by statusCode or res?
I'm trying to use `autoLogging` conditionally by `res.statusCode` without success.

It would be great if something like this would be available:

```typescript
import expressPino from 'express-pino-logger';

export const expressPinoLogger = expressPino({
  autoLogging: {
    ignore: (req, res) => {
       return res.statusCode !== 200;
    }
  }
});
```

filoscoder:
### Description

This PR begin from a specific need:
 - Love autoLogging
 - Love serializers
 - Doesn't need health check logs
 - Only needs logs for debug
 - `Can I prevent autoLogging by Response status code?` **NO**
 
I just add an early return in the `onResFinished` method. When the `level`, coming either from `useLevel` or `customLogLevel` option, the method finishes early without calling the logger.

This PR doesn't have breaking changes.

### Related Issue

resolve #190

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ $ node example.js | pino-pretty
 * `logger`: parent pino instance for a child logger instance, which will be used by `pino-http`. To refer to this child instance, use [pinoHttp.logger](#pinohttplogger-plogger)
 * `genReqId`: you can pass a function which gets used to generate a request id. The first argument is the request itself. As fallback `pino-http` is just using an integer. This default might not be the desired behavior if you're running multiple instances of the app
 * `useLevel`: the logger level `pino-http` is using to log out the response. default: `info`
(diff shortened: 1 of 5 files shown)


filoscoder:
I need this asap, would you please give a review? @mcollina  @jsumners

mcollina (approved):
lgtm

jsumners (approved):
Looks good to me.

filoscoder:
Thanks for the review! @mcollina @jsumners 
It will take longer to merge? I need this to finish our server log system.
Btw I'm really impressed for the performance of pino 🏎💨

jsumners:
Please see my `README` -- https://github.com/jsumners

We will issue the release when one of us is available to do so.

filoscoder:
> Please see my `README` -- https://github.com/jsumners
> 
> We will issue the release when one of us is available to do so.

fair enough, but don't forget this pr please! 🤓

PR #197, opened 2022-01-13
Give customProps access to the response

travisbeck:
This PR fixes the issues with #168 and also fixes the regression it introduced where customProps wouldn't be set until the response was sent (if you were using customReceiveMessage or just any other logging on your server).

The downsides are the customProps gets called twice, but it feels worth it to me personally.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -5,6 +5,7 @@ const serializers = require('pino-std-serializers')
 const getCallerFile = require('get-caller-file')
 const URL = require('fast-url-parser')
 const startTime = Symbol('startTime')
+var reqObject = Symbol('reqObject')
 
 function pinoLogger (opts, stream) {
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):

mcollina (approved):
lgtm

Spence-S:
@mcollina @travisbeck  -- I don't understand why customProps child needs to be added in both the `onResFinished` function and the middleware function? - it causes any custom props to _always_ run into the duplicate child logs problem. -- I think this is problematic as a user who wants to both use customProps and avoid the duplicate keys problem. 

Since we are setting the child logger in the middleware - it should be the same instance in the `onResFinished` function and it shouldn't ever need to be set again, unless I am missing something!

Issue #209, opened 2022-04-08
pino should be a peer dependency


trevorr:
I think pino should be a [peer dependency](https://nodejs.org/es/blog/npm/peer-dependencies/) (and [also a dev dependency](https://betterprogramming.pub/understanding-dev-peer-and-regular-dependencies-in-javascript-f4090621c7a7#:~:text=you%20also%20must%20include%20it%20in%20your%20devDependencies%20section)) of pino-http. I'm trying to use a pre-release version of pino, but since pino-http has a direct dependency on pino, the latest release version is also being installed and hides the version I want to use (unless I exclude it from my package-lock.json).

mcollina:
peerdependencies do not work well across all package managers. Specifically npm v6 has a significantly different behavior than npm v8 and yarn.

trevorr:
That's a little vague. Are there specific problems or scenarios you're worried about? Peer dependencies were created in npm 1 for this exact scenario. The behavior did change a bit in npm 7+ (to automatically install peer dependencies if the consumer didn't), but that seems like a good thing. I'm having (temporary) difficulty using this package due to the direct dependency.

trevorr:
FWIW, peer dependencies are also widely used in the react ecosystem across many npm and yarn versions. I've never run into any issues there.

jsumners:
They cause tremendous issues when not installing as a hoisted flat `node_modules` tree. Even then, they cause difficult problems.

trevorr:
I'd love to learn from your experiences here. Can you be more specific? Googling for things like "peerDependencies problems" isn't yielding much.

jsumners:
Let's say you have a package `foo` that depends upon `bar` which itself has a peer dependency of `baz`. So you write `foo` such that it has direct dependencies on `bar` and `baz`. All is good, you ship.

Now you write an application that uses `foo`, but also uses `baz` in its test suite. Depending on how `bar` looks for `baz`, it's very likely that `bar` will not be able to find `baz` because it has been hoisted away from the right tree.

This is one real example.

alexedwardjones:
I'm having similar issues with `pino` not being a peer dependency. Using yarn pnp it just errors.

```
Error: nestjs-pino tried to access pino, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

jsumners:
Please file your issue with `nestjs-pino`. We do not maintain that library.

PR #210, opened 2022-04-08
Move pino from dependencies to peer/dev

trevorr:
Fixes https://github.com/pinojs/pino-http/issues/209

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,10 +4,12 @@
   "description": "High-speed HTTP logger for Node.js",
   "main": "logger.js",
   "types": "index.d.ts",
+  "peerDependencies": {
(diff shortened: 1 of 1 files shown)


mcollina (changes requested):
This still support npm v6, which won't install peerdependencies by default (among other things). `peerDependencies` have always caused more headaches than the problem they solved. Maybe we can reconsider when we dump Node v12/npm v6.

rbnayax:
Any news on the subject? This still hurts for pnpm like flat node module structure

mcollina:
I'm really against using peerDepedencies due to the constant problem they cause.

jsumners:
Same.

PR #223, opened 2022-06-07
created a branch with a better fileStructure

dejanero:
Improved the file structure of `pino-http`.

- Added a 'src' folder and moving the source code there
- Added a 'examples' folder and moving the examples there
- Renaming the 'test' folder to 'tests'
- Changed the target folder for `npm publish` to 'src' so that it only includes the files there.


jsumners:
Thank you for attempting to contribute, but we do not need this work done.

Issue #228, opened 2022-06-14
Remove PINOHTTP_DEP001 and PINOHTTP_DEP002


mcollina:
Their time is overdue and the deprecated functionality should be removed.

Puppo:
Hi @mcollina,
Can I help you with this?

mcollina:
Go for it!

vikaspotluri123:
@mcollina now that #229 is merged, can this be closed?

mcollina:
good spot, thanks!

PR #229, opened 2022-06-14
Remove PINOHTTP_DEP001 and PINOHTTP_DEP002

Puppo:
Remove PINOHTTP_DEP001 and PINOHTTP_DEP002 warning like describing in #228 
Remove references to `autoLogging.ignorePaths` and to `autoLogging.getPath`, from now they are replace definitely by `autoLogging.ignore`.


deprecations.js
```diff
@@ -1,11 +0,0 @@
-'use strict'
```
jsumners, line 1: Please keep the file. We can comment out the two warning lines. We will want the file for future deprecations.

jsumners (changes requested):

deprecations.js
Puppo: @jsumners ok, I restored the file and I commented the warning name and an example for the future deprecations.

deprecations.js
```diff
@@ -2,10 +2,8 @@
...
 const warning = require('process-warning')()
 
-const warnName = 'PinoHttpWarning'
-
-warning.create(warnName, 'PINOHTTP_DEP001', 'Use of "autoLogging.ignorePaths" is deprecated since version 7.1.0, use "opts.autoLogging.ignore" instead.')
-
-warning.create(warnName, 'PINOHTTP_DEP002', 'Use of "autoLogging.getPath" is deprecated since version 7.1.0, use "opts.autoLogging.ignore" instead.')
+// Warning example
+// const warnName = 'PinoHttpWarning'
+// warning.create(warnName, 'PINOHTTP_DEP001', 'Use of "autoLogging.ignorePaths" is deprecated since version 7.1.0, use "opts.autoLogging.ignore" instead.')
```
jsumners, line 7: Please keep both warnings around so that we know what number we are on. As written, I am quite certain that the next time we add a deprecation it will get tagged as `PINOHTTP_DEP002`.

Puppo: ok, done!

jsumners (approved):

mcollina (approved):
lgtm

PR #231, opened 2022-06-20
Add custom object hooks

Simon-Campbell:
This adds a functionality for overriding/customising the structured object logged at request received/success/error. It is *like* the custom[Received|Success|Error]Message hooks, but works on customising the structured object.

It was opted not to overload *Message due to that being specifically about messages, and the risk that adjusting that could break an existing usage of that hook.

The README has been updated with example usage.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -6,30 +6,6 @@ const getCallerFile = require('get-caller-file')
 const startTime = Symbol('startTime')
 const reqObject = Symbol('reqObject')
 
-function GetFunctionOrDefault (value, defaultValue) {
-  if (value && typeof value === 'function') {
-    return value
-  }
-
-  return defaultValue
-}
-
-function DefaultSuccessfulRequestObjectProvider (req, res, successObject) {
-  return successObject
-}
-
-function DefaultFailedRequestObjectProvider (req, res, error, errorObject) {
-  return errorObject
-}
-
-function DefaultFailedRequestMessageProvider () {
-  return 'request errored'
-}
-
-function DefaultSuccessfulRequestMessageProvider (req, res) {
-  return res.writableEnded ? 'request completed' : 'request aborted'
-}
-
 function pinoLogger (opts, stream) {
   if (opts && opts._writableState) {
     stream = opts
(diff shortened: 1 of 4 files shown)


mcollina (changes requested):

Simon-Campbell:
> Thanks for opening a PR! Can you please add a unit test?

Thanks for feedback. I have now added tests as an extra commit.

logger.js
```diff
@@ -64,8 +64,13 @@ function pinoLogger (opts, stream) {
   const autoLoggingIgnore = opts.autoLogging && opts.autoLogging.ignore ? opts.autoLogging.ignore : null
   delete opts.autoLogging
 
+  const onRequestReceivedObject = opts.customReceivedObject && typeof opts.customReceivedObject === 'function' ? opts.customReceivedObject : undefined
```
mcollina, line 67: Could you move the function to be a top-level function instead of a closure? I think the code would read better.

logger.js
```diff
@@ -147,9 +153,20 @@ function pinoLogger (opts, stream) {
...
+          const receivedObjectResult = onRequestReceivedObject !== undefined ? onRequestReceivedObject(req, res, undefined) : undefined
+          const receivedStringResult = receivedMessage !== undefined ? receivedMessage(req, res) : undefined
+
+          if (receivedObjectResult && receivedStringResult) {
+            req.log[level](receivedObjectResult, receivedStringResult)
+          } else if (receivedStringResult) {
+            req.log[level](receivedStringResult)
+          } else {
+            req.log[level](receivedObjectResult)
+          }
```
mcollina, line 169: There are quite a a lot of conditionals in this block. Could you simplify them a bit and remove some duplications?

mcollina (changes requested):

Simon-Campbell:
Hi @mcollina, 

I have addressed your feedback. I have also converted the common repeated pattern of (since I believe it further simplified what was being expressed.)

`const someEffectiveValue = opt.someOption && opt.someOption === 'function' ? opt.someOption : SomeDefaultOption;
`

to:

`
const someEffectiveValue = getFunctionOrDefault(opt.someOption, SomeDefaultOption);`

mcollina (approved):
lgtm

PR #239, opened 2022-07-13
refactor(logger): remove uneeded defensive code

Fdawgs:
`autoLoggingIgnore` is already checked to not be null on the line above:

https://github.com/pinojs/pino-http/blob/5d8a2f5a57c7ce52e4170ecf0e8d41e5b9fadb38/logger.js#L157


mcollina (approved):
lgtm

PR #240, opened 2022-07-27
Pass response object to genReqId()

bmodotdev:
Passing the response body object to genReqId() allows us to more easily
share the request ID throughout our code, e.g. setting an “X-Request-Id”
header via “response.header()” or storing it in “response.locals” for
rendering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,13 +122,21 @@ $ node example.js | pino-pretty
 
 const http = require('http')
 const server = http.createServer(handle)
+const { v4: uuidv4 } = require('uuid')
 const pino = require('pino')
 const logger = require('pino-http')({
   // Reuse an existing logger instance
   logger: pino(),
 
   // Define a custom request id function
-  genReqId: function (req) { return req.id },
+  genReqId: function (req, res) {
+    if (req.id) return req.id
+    let id = req.get('X-Request-Id')
+    if (id) return id
+    id = uuidv4()
+    res.header('X-Request-Id', id)
+    return id
+  },
 
   // Define custom serializers
   serializers: {
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ $ node example.js | pino-pretty
 
 const http = require('http')
 const server = http.createServer(handle)
-const { v4: uuidv4 } = require('uuid')
+const { randomUUID } = require('node:crypto')
 const pino = require('pino')
 const logger = require('pino-http')({
   // Reuse an existing logger instance
@@ -133,7 +133,7 @@ const logger = require('pino-http')({
     if (req.id) return req.id
     let id = req.get('X-Request-Id')
     if (id) return id
-    id = uuidv4()
+    id = randomUUID()
     res.header('X-Request-Id', id)
     return id
   },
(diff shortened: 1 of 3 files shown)


bmodotdev:
This feels a like silly commit given its size, but I believe providing the response object makes a lot of sense. I can see `req` makes sense when the client is querying the server with an already existing `X-Request-Id` header, so you simply take it from the `req` object.

However, in the case that a request id does not exist in the `req` object, if you generate one here in your logger, you must store it in the `req` object for later reference, then later pull it out for setting the header.

To workaround this, I just created a UUID middleware that generates the id, sets the header, then stores it in `req` for exposing to the logger via `genReqId()`.

Ideally, I would not need a middleware, `genReqId()` would simply pass the `req` and `res` objects and I simply generate it there, set it in headers, and return it to the logger, all in one place.

Anyway, just a small improvement. Thanks for the awesome library!

mcollina (changes requested):

bmodotdev:
> Thanks for opening a PR! Can you please add a unit test?

I am having a hard time thinking of how a unit test for passing express response object might function to `genReqId()`. Passing the response object into `genReqId()` is moreso for the benefit of other code, i.e. setting a header or setting `res.locals`. I have updated the README though, and updated the unit tests to pass `res`.

If that is not sufficient, should I just use `res` to set a header or set a `res.locals.id`, then check it? Sorry, still understanding the codebase.

README.md
```diff
@@ -122,13 +122,21 @@ $ node example.js | pino-pretty
 
 const http = require('http')
 const server = http.createServer(handle)
+const { v4: uuidv4 } = require('uuid')
```
mcollina, line 125: Could you recommend https://nodejs.org/api/crypto.html#cryptorandomuuidoptions instead? It's _faster_ and does not require external deps.

bmodotdev: absolutely, thanks for the suggestion, this is done

test/test.js
```diff
@@ -290,7 +290,7 @@ test('uses a custom genReqId function', function (t) {
 
   const dest = split(JSON.parse)
   let idToTest
-  function genReqId (req) {
+  function genReqId (req, res) {
```
mcollina, line 293: Please add an assertion to verify that `res` is set.

test/test.js
```diff
@@ -536,7 +536,7 @@ test('support a custom instance with custom genReqId function', function (t) {
   const dest = split(JSON.parse)
 
   let idToTest
-  function genReqId (req) {
+  function genReqId (req, res) {
```
mcollina, line 539: Please add an assertion to verify that `res` is set.

mcollina (changes requested):

test/test.js
bmodotdev: This is done

test/test.js
bmodotdev: This is done

mcollina (approved):
lgtm

bmodotdev:
Awesome, thanks for the guidance. That simple assertion check makes much more since now, I was trying to over complicate it :)

MrTalon63:
GenReqId interface does not have `res` specified resulting in typescript error

PR #256, opened 2022-11-14
feat: add option `quietResLogger` to supress response data from res.log entries

### LINKED ISSUE #235: [Proposal] support for quietResLogger (similar to quietReqLogger) - log full bindings at start of request instead at the end.
It's currently hard to implement following behavior with pino-http:

1. Log request **with headers, url, request params and other required info.**
2. Process request and call `log` multiple times during the scope of request. **Request params, headers, etc.. should NOT be included in these logs.**
3. Log response. **Request params, headers, etc... should NOT be added to this log.**

It seems like with current functionality we can either add unnecessary metadata to every single log during scope of request (`quietReqLogger=false`) or add them at the end (`quietReqLogger=true`).
However, there is no way to easily do the opposite - to add all aditional info at start of request, and do only lightweight logging afterwards. It seems natural, that first log of request scope should include all the metadata (i..e what if request takes way longer than expected or it hangs somehow).

micalevisk:
closes #235 

I didn't succeeded on making the `npm t` running with no errors, even in the fresh new cloned repo :thinking:

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -77,6 +77,7 @@ function pinoLogger (opts, stream) {
   delete opts.customErroredMessage
 
   const quietReqLogger = !!opts.quietReqLogger
+  const quietResLogger = !!opts.quietResLogger
 
   const logger = wrapChild(opts, theStream)
 
@@ -114,7 +115,6 @@ function pinoLogger (opts, stream) {
 
       log[level](
         onRequestErrorObject(req, this, error, {
-          [resKey]: this,
           [errKey]: error,
           [responseTimeKey]: responseTime
         }),
@@ -126,7 +126,6 @@ function pinoLogger (opts, stream) {
 
     log[level](
       onRequestSuccessObject(req, this, {
-        [resKey]: this,
         [responseTimeKey]: responseTime
       }),
       successMessage(req, this, responseTime)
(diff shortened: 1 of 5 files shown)


mcollina (changes requested):
can you add the docs to the README?

logger.js
```diff
@@ -138,16 +137,20 @@ function pinoLogger (opts, stream) {
 
     req.id = genReqId(req, res)
 
-    const log = quietReqLogger ? logger.child({ [requestIdKey]: req.id }) : logger
+    const reqLog = quietReqLogger ? logger.child({ [requestIdKey]: req.id }) : logger
+    let fullReqLogger = reqLog.child({ [reqKey]: req })
+
+    const resLog = quietResLogger ? logger.child({ [requestIdKey]: req.id }) : logger
+    let fullResLogger = resLog.child({ [resKey]: res })
```
micalevisk, line 144: > What errors are you getting?

@mcollina a bunch of tests stop passing after adding this line

mcollina: The tests should be updated

vorticalbox: could we not do this?

```
const log = quietReqLogger  || quietResLogger ? logger.child({ [requestIdKey]: req.id }) : logger
```

and leave the rest of the code as is?

micalevisk: hm, I just did it but the tests are still not passing 

image

vorticalbox: I actually really want this feature, I'll pull tomorrow and use some work time on it

vorticalbox:
I have been playing and have it successfully passing tests and blocking the response log.

however i am getting this error

image

vorticalbox:
actually the misunderstood the the ticket, I wanted a way to disable the request complete log. have a PR here https://github.com/pinojs/pino-http/pull/259

Issue #260, opened 2022-11-29
Multiple middleware instances with different redact config


mmadson:
- Create 2 pinoHttp loggers with different redact configs. 
- Attach both middleware to express app

Expected: 2 log entries with different redactions
Actual: 2 log entries with the same redactions

I think this is because the logger variable of the loggingMiddleware function is shared across multiple instantiations of the middleware. It would be good if we can have separate logger instances for each call of pinoHttp

jsumners:
It seems like you are passing `opts = { logger: pino() }` to both instances. Otherwise, a unique instance is created each time at https://github.com/pinojs/pino-http/blob/112e2eb2297e30516248684a29ee0658a5cc8b30/logger.js#L202

mmadson:
That was my expectation as well, however my config does not specify a pino logger in the opts:

```
VERY_REDACTED_HTTP_LOGGER = pinoHttp({
    genReqId: (req: express.Request) => {
      if (!req.id) {
        req.id = randomUUID();
      }
      return req.id;
    },
    customProps: (_req: IncomingMessage, _res: ServerResponse) => {
      return deepmerge(getLogContext(), { requestLog: false });
    },
    customLogLevel: (_req: IncomingMessage, res: ServerResponse, _error: Error) => (res.statusCode >= 500 ? "error" : "info"),
    redact: config.logger.redacted,
  } as Options);
  
  SOMEWHAT_REDACTED_LOG_HTTP_LOGGER = pinoHttp({
    genReqId: (req: express.Request) => {
      if (!req.id) {
        req.id = randomUUID();
      }
      return req.id;
    },
    customProps: (_req: IncomingMessage, _res: ServerResponse) => {
      return deepmerge(getLogContext(), { requestLog: true });
    },
    customLogLevel: (_req: IncomingMessage, res: ServerResponse, _error: Error) => (res.statusCode >= 500 ? "error" : "info"),
    redact: config.logger.redacted.filter((key) => key !== "req.url"),
  } as Options);
  ```
  
  I think the culprit is this line:
  
  https://github.com/pinojs/pino-http/blob/master/logger.js#L88
  
Which assigns the logger (either a new logger or the opts logger) to a property of the loggingMiddleware function which if I'm not mistaken is shared between multiple pino-http middlewares.

jsumners:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

mmadson:
[...]

Issue #30, opened 2017-04-19
Logging req.user properties


scott-martin:
I would really like to log some properties off of `req.user` which is added in middleware that is after pino-http. It looks like pino-http serializes the `req` object immediately because it creates a child logger with the request object. Would it be possible to postpone serialization of req until the request finishes? Or possibly add another hook that could be used to serialize once the request is finished? Could anyone offer a suggestion of how I could accomplish what I'm trying to do here?

davidmarkclements:
Yep totally possible - 

You need to do two things

1. place the logging middleware beneath the middleware that adds `req.user`
2. pass a custom serializer



```
const pino = require('pino')()
const pinoMiddleware = require('pino-http') // or require('express-pino-logger') or require('koa-pino-logger) - whatever it is you're using, concepts are the same

app.use(middlewareWhichAddsUserToReq)

app.use(pinoMiddleware({
  serializers: {
    req: function customReqSerializer (req) {
      return {
        id: req.id,
        method: req.method,
        url: req.url,
        user: req.user,
        headers: req.headers,
        remoteAddress: req.connection.remoteAddress,
        remotePort: req.connection.remotePort
      }
   }
 }
})
```

Closing for now, but let me know if that works for you :bowtie:

scott-martin:
Would this still log the request if the authentication middleware threw an unauthorized exception?

scott-martin:
Yeah I just verified that it does not in fact log the request if the authentication middleware fails. Here is a workaround that I was able to do to get it working.

```
const pino = require('pino')();
const pinoMiddleware = require('pino-http');

app.use(pinoMiddleware); // ensures we capture all requests
app.use(authenticationMiddleware); // adds `req.user`
app.use((req, res, next) => {
   if (req.user) {
        req.log = res.log = req.log.child({ 
            user: req.user
        });
    }
    next();
}); // redefine the res.log object as a new child logger that includes the user object
```

This works because in line 32 of logger.js you can see that pino-http is using the logger object that is defined on `res` to perform the final log.

davidmarkclements:
hmm that does mean you're creating a child of a child logger, although the perf hit is still probably negligible. Another way is to use a getter: 

```js
app.use(pinoMiddleware({
  serializers: {
    req: function customReqSerializer (req) {
      return {
        id: req.id,
        method: req.method,
        url: req.url,
        get user () { return req.user },
        headers: req.headers,
        remoteAddress: req.connection.remoteAddress,
[...]

PR #263, opened 2022-12-07
Feature: Allow multiple instances of middleware

mmadson:
PR to address Issue #260 

- scope pino loggers to lambda closures instead of function properties
- set request and response log property to first middleware pino logger but capture additional middleware pino loggers in new allLogs request and response property
- unit tests to confirm each middleware redact properties are independent

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: CI
 
 on:
+  workflow_dispatch:
   push:
     paths-ignore:
         - 'docs/**'
--- a/index.d.ts
+++ b/index.d.ts
@@ -80,6 +80,7 @@ declare module "http" {
     interface IncomingMessage {
       id: ReqId;
       log: pino.Logger;
+      allLogs: pino.Logger[];
     }
 
     interface ServerResponse {
--- a/index.test-d.ts
+++ b/index.test-d.ts
@@ -18,7 +18,6 @@ pinoHttp({ genReqId: (req: IncomingMessage, res: ServerResponse) => res.statusCo
 pinoHttp({ genReqId: (req: IncomingMessage, res: ServerResponse) => 'foo' });
 pinoHttp({ genReqId: (req: IncomingMessage, res: ServerResponse) => Buffer.allocUnsafe(16) });
 pinoHttp({ useLevel: 'error' });
-pinoHttp({ prettyPrint: true }); // deprecated but still present in pino.
 pinoHttp({ transport: { target: 'pino-pretty', options: { colorize: true } } });
 pinoHttp({ autoLogging: false });
 pinoHttp({ autoLogging: { ignore: (req: IncomingMessage) => req.headers['user-agent'] === 'ELB-HealthChecker/2.0' } });
--- a/index.test-d.ts
+++ b/index.test-d.ts
@@ -18,7 +18,6 @@ pinoHttp({ genReqId: (req: IncomingMessage, res: ServerResponse) => res.statusCo
 pinoHttp({ genReqId: (req: IncomingMessage, res: ServerResponse) => 'foo' });
 pinoHttp({ genReqId: (req: IncomingMessage, res: ServerResponse) => Buffer.allocUnsafe(16) });
 pinoHttp({ useLevel: 'error' });
-pinoHttp({ prettyPrint: true }); // deprecated but still present in pino.
 pinoHttp({ transport: { target: 'pino-pretty', options: { colorize: true } } });
 pinoHttp({ autoLogging: false });
 pinoHttp({ autoLogging: { ignore: (req: IncomingMessage) => req.headers['user-agent'] === 'ELB-HealthChecker/2.0' } });
@@ -157,6 +156,7 @@ const stdSerializedResults: StdSerializedResults = {
 const httpServerListener: RequestListener = (request, response) => {
   // req.log and req.id should be available
   request.log.info(`Request received with request ID ${request.id}`);
+  request.allLogs[0].info("Request Received");
   // res[startTime] should be available
   response[startTime] = Date.now();
   response.end("Hello world");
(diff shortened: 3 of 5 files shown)


index.test-d.ts
```diff
@@ -18,7 +18,6 @@ pinoHttp({ genReqId: (req: IncomingMessage, res: ServerResponse) => res.statusCo
 pinoHttp({ genReqId: (req: IncomingMessage, res: ServerResponse) => 'foo' });
 pinoHttp({ genReqId: (req: IncomingMessage, res: ServerResponse) => Buffer.allocUnsafe(16) });
 pinoHttp({ useLevel: 'error' });
-pinoHttp({ prettyPrint: true }); // deprecated but still present in pino.
```
mmadson, line 21: git hooks, tests and compilation failed unless I removed this line; not sure still required and still present in upstream pino

index.d.ts
```diff
@@ -80,6 +80,7 @@ declare module "http" {
     interface IncomingMessage {
       id: ReqId;
       log: pino.Logger;
+      allLogs: pino.Logger[];
```
mmadson, line 83: when multiple pino-http middleware are registered, the first registered middleware's pino logger will be attached to the req.log property. The remaining middleware's pino loggers (as well as the first) will be available in this new allLogs property.

logger.js
```diff
@@ -85,58 +85,56 @@ function pinoLogger (opts, stream) {
...
 
   const genReqId = reqIdGenFactory(opts.genReqId)
-  loggingMiddleware.logger = logger
-  return loggingMiddleware
-
-  function onResFinished (err) {
-    this.removeListener('close', onResFinished)
-    this.removeListener('error', onResFinished)
-    this.removeListener('finish', onResFinished)
+  const result = (req, res, next) => {
```
mmadson, line 88: instead of attaching the pino logger to the statically defined loggingMiddleware function and returning the single reference to it, we create a new closure called result and attach the pino logger to the result. Each call to pinoHttp will create a new result closure so each call gets it's own closure rather than sharing the loggingMiddleware function and it's logging property.

logger.js
```diff
@@ -85,58 +85,56 @@ function pinoLogger (opts, stream) {
...
+      successMessage(req, res, responseTime)
     )
   }
 
-  function loggingMiddleware (req, res, next) {
+  function loggingMiddleware (logger, req, res, next) {
     let shouldLogSuccess = true
 
-    req.id = genReqId(req, res)
+    req.id = req.id || genReqId(req, res)
```
mmadson, line 137: when multiple middleware are registered we have to decide how the req.id is to be set. Convention established here is that the first middleware's genReqId function will set the req.id, subsequent middleware's genReqId functions will be ignored, so the req.id will be set only once per request. 

This may warrant additional discussion.

logger.js
```diff
@@ -85,58 +85,56 @@ function pinoLogger (opts, stream) {
...
+        [resKey]: res,
         [responseTimeKey]: responseTime
       }),
-      successMessage(req, this, responseTime)
+      successMessage(req, res, responseTime)
     )
   }
 
-  function loggingMiddleware (req, res, next) {
+  function loggingMiddleware (logger, req, res, next) {
```
mmadson, line 134: since the logger is attached to the result closure instead of this function, we pass the logger into the function as a param instead of assuming it is attached as function property.

logger.js
```diff
@@ -85,58 +85,56 @@ function pinoLogger (opts, stream) {
...
-      log = this.log.child(customPropBindings)
+      log = logger.child(customPropBindings)
     }
 
-    if (err || this.err || this.statusCode >= 500) {
-      const error = err || this.err || new Error('failed with status code ' + this.statusCode)
+    if (err || res.err || res.statusCode >= 500) {
+      const error = err || res.err || new Error('failed with status code ' + res.statusCode)
 
       log[level](
```
mmadson, line 113: the current implementation emits the final log message on the single logger attached to the response object. Instead we have passed the intended logger in as a parameter from a new closure onResponseComplete so that each middleware generates it's logs to its own pino logger, not relying on the logger attached to the response object.

mcollina (approved):
lgtm

mmadson:
> lgtm

Looks like CICD had some tests that either don't run by default locally or at least aren't run the same way and they didn't like my usage of the nullish coalescing operator `??=`. I went ahead and replaced it with and if check to instantiate the allLogs array. Hopefully that gets us past the CICD checks. If there is a way for me to run them locally just let me know and I'll try to resolve any other failures before pushing more code. As far as I can tell `yarn test` passes for me locally.

mmadson:
@mcollina Whenever you have a free moment would you mind kicking off the workflow again? I don't seem to have permission but hoping that my fix has addressed the CICD issue. 🙏

mmadson:
@mcollina Oof; okay so what is the saying? Third time is the charm? I was worried that perhaps my removal of the prettyPrint property typescript test may have been premature and although it wasn't passing for me locally perhaps it would pass on CICD -- so I restored the line while fixing the `??=` operator compiler error. However, it looks like the prettyPrint issue is not only an issue for me locally but also an issue on CICD so I went ahead and removed it again. I think now everything should be good to go and hopefully the next run will be a clean CICD run. 🤞

.github/workflows/ci.yml
```diff
@@ -1,6 +1,7 @@
 name: CI
 
 on:
+  workflow_dispatch:
```
mmadson, line 4: This allows you to manually run your GItHub CICD workflow from the Actions tab without pushing a commit. If you would prefer me to revert this change I can. Just LMK.

mmadson:
@mcollina I just realized that I can enable the same GitHub Actions CICD pipeline on my fork. So I went ahead and did that and found another `??=` that I needed to replace and verified that the CICD all passes on my forked repo

image

So feel free to start the CICD workflow whenever you are ready we should be green. =)

mmadson:
@mcollina Latest failure seems to be a problem with how the test was written. The test assumes the sample server will always result in a 0 latency response time but in this particular test run. The response time turned out to be 1. I'll go ahead and update the test to match any digits instead of just 0 which should make the test more robust. 

```
            --- expected
            +++ actual
            @@ -1,1 +1,1 @@
            -Response time is: 0 GET
            +Response time is: 1 GET
```

mmadson:
Yay, finally an all green CICD run. Guess it's up to y'all (@mcollina) to decide when to merge and cut a release. Thanks again for releasing this project to OSS, it's proven very beneficial for our request / response logging. 🎉

kohtala:
People modify the req.log to add properties into log. #30 With this change, the properties are no longer logged.

PR #265, opened 2022-12-14
fix: set error param as optional for options.customLogLevel

raiman264:
according to API docs the error param in customLogLevel function is optional
but that is not reflected in the typescript definition

> customLogLevel: set to a function (req, res, err) => { /* returns level name string */ }. This function will be invoked to determine the level at which the log should be issued (silent will prevent logging). This option is mutually exclusive with the useLevel option. The first two arguments are the HTTP request and response. **The third argument is an error object if an error has occurred in the request**.


mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test? We use tsd for typescript.

mcollina (approved):
lgtm

PR #269, opened 2023-01-25
Switch request id generator to a random string

marcbachmann:
### Motivation
The current integer-based request id generation is not unique enough when running multiple processes or serving many requests. This prompted implementors to always configure it. The new string based random id generator should serve as production-ready config.

### Implementation
The id generation uses the concept of https://github.com/mcollina/hyperid but does not introduce any dependencies as the logic is rather simple.

It will generate ids from `UCGtioQAmusXkiWj-1` to `UCGtioQAmusXkiWj-9999`.
Therefore have a length from 18 to 21 characters, where the random prefix is 16 characters.

### Benchmark
With a microbenchmark it turns out 12 bytes and a small counter provided the fastest execution time on all node versions.
Much higher numbers like Number.MAX_SAFE_INTEGER as `count` also resulted in slower computations (twice as slow, but it's probably neglectable anyways).

12 bytes result in 16 characters as prefix, which should provide plenty of uniqueness according to https://alex7kom.github.io/nano-nanoid-cc

```
Benchmarking 1'000'000 operations per run.
---
Benchmark: current genReqId
finished 100 runs (0.2s)
Average per run: 2.0ms
---
Benchmark: crypto.randomUUID
finished 100 runs (7.8s)
Average per run: 78.1ms
---
Benchmark: new genReqId with eight bytes
finished 100 runs (1.8s)
Average per run: 18.2ms
---
Benchmark: new genReqId with twelve bytes (the version in this PR)
finished 100 runs (1.2s)
Average per run: 12.4ms
```

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -228,10 +228,18 @@ function wrapChild (opts, stream) {
 
 function reqIdGenFactory (func) {
   if (typeof func === 'function') return func
-  const maxInt = 2147483647
-  let nextReqId = 0
-  return function genReqId (req, res) {
-    return req.id || (nextReqId = (nextReqId + 1) & maxInt)
+  return defaultReqIdGenFactory()
+}
+
+function defaultReqIdGenFactory () {
+  const { randomFillSync } = require('crypto')
+  const buffer = Buffer.alloc(12)
+  let id = randomFillSync(buffer).toString('base64url')
+  let count = 0
+
+  return function genReqId () {
+    if (count++ < 9999) return `${id}-${count}`
+    return `${(id = randomFillSync(buffer).toString('base64url'))}-${(count = 1)}`
   }
 }
 
@@ -268,3 +276,4 @@ module.exports.stdSerializers = {
 module.exports.startTime = startTime
 module.exports.default = pinoLogger
 module.exports.pinoHttp = pinoLogger
+module.exports.defaultReqIdGenFactory = defaultReqIdGenFactory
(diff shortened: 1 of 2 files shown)


logger.js
```diff
@@ -268,3 +276,4 @@ module.exports.stdSerializers = {
 module.exports.startTime = startTime
 module.exports.default = pinoLogger
 module.exports.pinoHttp = pinoLogger
+module.exports.defaultReqIdGenFactory = defaultReqIdGenFactory
```
marcbachmann, line 279: This should probably exposed to the tests 🤔 
Maybe using a symbol or separate file?

mcollina:
I would prefer to not change the default for something that is 6x slower :(. We should see if this is a bottleneck for actual servers.

jsumners:
Provide it as an optional default, e.g. https://getpino.io/#/docs/api?id=pino-stdtimefunctions. 

Use `hyperid` as a dependency if that is what is desired. I know that it handles `> 9999` in some way.

Put it in something like `lib/req-id-hyperid.js` and test it individually.

marcbachmann:
> Provide it as an optional default, e.g. https://getpino.io/#/docs/api?id=pino-stdtimefunctions.

It's not really a stdtimefunction. so we'd need a new config for those 🥲 

> Use hyperid as a dependency if that is what is desired. I know that it handles > 9999 in some way.

Hyperid is 4 times slower than the new function 😉 

I'll close this for now as people could just declare it using
```js
pino({
  genReqId: require('hyperid')()
})
```

PR #253, opened 2022-10-02
Add success response time

baterson:
Fix https://github.com/pinojs/pino-http/issues/199
Please also add "hacktoberfest-accepted" label :)

PR #270, opened 2023-01-29
Fix customSuccessMessage ts definition

robmcguinness:
https://github.com/pinojs/pino-http/pull/253 added `responseTime` value to be passed into the `customSuccessMessage` callback but the TS definition file was not updated.


mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test? For typescript we use tsd.

robmcguinness:
Whoops will do.

robmcguinness:
Added. Hopefully correct.

mcollina (approved):
lgtm

PR #271, opened 2023-01-31
docs(readme): replace shields with native badge

Fdawgs:
See https://github.com/badges/shields/issues/8671, resolves badge not rendering


jsumners:
https://github.com/pinojs/pino/pull/1636#pullrequestreview-1277579641

PR #272, opened 2023-01-31
Fix customLogLevel usage in onResFinished function

kanmir:
Hi!

I found a bug.
There is a missing `req` argument in `getLogLevelFromCustomLogLevel` function call [at onResFinished](https://github.com/pinojs/pino-http/blob/master/logger.js#L97)

Because of this, I got an error `TypeError: Cannot read properties of undefined (reading 'req')`, when using `customLogLevel` option


mcollina (changes requested):

mcollina (approved):
lgtm

PR #276, opened 2023-02-28
docs(readme): fix genReqId() example

GabenGar:
The example in the docs assumes `expressjs` `req` and `res` arguments, but the types in the library assume vanilla `req` and `res`. There is no mention of minimum required NodeJS version so I assume it's at least 14+, which supports [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing#browser_compatibility).

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,11 +130,10 @@ const logger = require('pino-http')({
 
   // Define a custom request id function
   genReqId: function (req, res) {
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

Issue #216, opened 2022-05-05
customProps always logs duplicate keys


Spence-S:
I am working with google cloud logging and it doesn't play nicely with the duplicate keys:

```js
    customProps(req) {
      // adds a custom object to all http logs that stackdriver
      // uses to make logs more informative and visible at a glance
      // see: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#httprequest
      // for the details of the structured log messages
      return {
        httpRequest: {
          requestMethod: req.method,
          requestUrl: req.url,
          userAgent: req.headers['user-agent'],
          remoteIp:
            req.headers['x-forwarded-for']?.split(',').shift() ||
            req.socket?.remoteAddress,
        },
      };
    },
```

For instance:
custom props cause the following log:
`"{ "httpRequest": {"method":"GET"},"httpRequest": {"method":"GET"}}"`

ends up in stackdriver/google cloud logging looking like:
`{httpRequest: { method: 'GETGET' }}`

mcollina:
Can you include what you are trying to achieve? The snippet is not enough.

Spence-S:
- I want to use `customProps` to add the httpRequest object for google cloud logging
- I want to avoid the [duplicate keys problem](https://github.com/pinojs/pino/blob/master/docs/child-loggers.md#duplicate-keys-caveat) 
- I do NOT want to use a transport to format or clean up the keys

Currently, any usage of `customProps` causes duplicate keys of whatever is added there because of this PR. https://github.com/pinojs/pino-http/pull/197

Right now, as a user, I am forced to deal with duplicate keys if I want to use the custom props option.

mcollina:
Your usecase is a bit far from mine.

Would you like to send a Pull Request to address this issue? Remember to add unit tests.

Spence-S:
I took a look at this over the weekend, and unfortunately I couldn't come up with a quick solution that doesn't cause unit tests to break. This one may require some time/care to fix correctly and maintain all current behavior. 

In order to pass the correct (final) res.statusCode to the customProps bindings and have the bindings in logs on `req.log` and `res.log` we do need to have the bindings applied in both places.

jamyouss:
I have the same issue.

Couldn't we just remove the `customPropBindings` on the `loggingMiddleware`

>  const customPropBindings = (typeof customProps === 'function') ? customProps(req, res) : customProps
    if (customPropBindings) {
[...]

PR #288, opened 2023-06-11
fix: twice customProps

youngkiu:
Hello!
I am the author of [pino-web-hook](https://getpino.io/#/docs/transports?id=pino-slack-webhook).

I am using the NestJS framework, and I wanted to display the `pinoHttp` property, thinking that the main function of `nestjs-pino` is Middleware that connects `pino-http`.
```
    LoggerModule.forRoot({
      pinoHttp: {
        customProps: (req, res) => ({
          context: 'HTTP',
        }),
      },
    }),
```
However, I found the following bug, and also found out that there is an issue https://github.com/pinojs/pino-http/issues/216.
image

I tried the following to solve this problem.
1. I tried to use [onchild-function](https://github.com/pinojs/pino/blob/master/docs/api.md#onchild-function), but stopped because the instance given as a parameter was a logger instance that could be known outside of OnChildCallback.
  ```js
+   opts.onChild = (instance) => {
+     // Execute call back code for each newly created child.
+   }
      const logger = wrapChild(opts, theStream)
  ```
2. The second method is the current PR. Additional conditional statements can be written on one line, but for readability, I have declared local variables.

I debugged it with the following modified [example.js](https://github.com/pinojs/pino-http/blob/master/example.js) file.
```js
const logger = require('./')({
  customProps: (req, res) => ({
    context: 'HTTP'
  })
})
```

Thanks

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const pino = require('pino')
+const { pino, symbols: { stringifySym, chindingsSym } } = require('pino')
 const serializers = require('pino-std-serializers')
 const getCallerFile = require('get-caller-file')
 const startTime = Symbol('startTime')
@@ -103,7 +103,11 @@ function pinoLogger (opts, stream) {
 
     const customPropBindings = (typeof customProps === 'function') ? customProps(req, res) : customProps
     if (customPropBindings) {
-      log = logger.child(customPropBindings)
+      const customPropBindingStr = logger[stringifySym](customPropBindings).replace(/[{}]/g, '')
+      const customPropBindingsStr = logger[chindingsSym]
+      if (!customPropBindingsStr.includes(customPropBindingStr)) {
+        log = logger.child(customPropBindings)
+      }
     }
 
     if (err || res.err || res.statusCode >= 500) {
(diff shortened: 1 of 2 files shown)


jmealo:
@youngkiu I thought this was by design to be fast... when the JSON is parsed, most parsers I have tested will keep the last entry logged, so there shouldn't be any dupes appearing in your log viewer?

youngkiu:
@jmealo Thanks for your interest in my fix.

I understand that using pino-pretty will output customProps without duplication due to JSON parsing.
Especially in the production log, I think it's more important to save it raw than to look pretty. - We plan to write a program that can analyze the accumulated logs if necessary in the future.
I don't use a separate JSON parser, but I modified it because duplicate logs were inconvenient.

mcollina (changes requested):

jmealo:
@youngkiu: If you need a hand with unit tests let me know. I think a failing test for the current implementation would be the easiest way to approach this.

youngkiu:
@jmealo I'm still not sure how I was able to see `logger[chindingsSym]` in unit test. Can you help me write the test code?

youngkiu:
@mcollina 
In previous commit, customProps is number 2,
From my fix, I added a test case to ensure that customProps is number 1.
Please review PR

mcollina (approved):
lgtm

jsumners (approved):

PR #290, opened 2023-07-21
Fix: #289 add documentation and typescript definitions for res.log

### LINKED ISSUE #289: Documentation: res.log() is only mentioned in quietReqLogger?
Hello,

I was reading the API docs, and noticed that `res.log()` is only mentioned in `quietReqLogger`:

> **quietReqLogger**: when true, the child logger available on req.log will no longer contain the full bindings and will now only have the request id bound at reqId (note: the autoLogging messages and the logger available on **`res.log`** will remain the same except they will also have the additional reqId property). default: false

Should we document `res.log()` elsewhere -- otherwise, it may give the impression of being a typo.

Thanks for the great work!

jmealo:
Fixes #289

### DIFF
--- a/index.test-d.ts
+++ b/index.test-d.ts
@@ -138,6 +138,8 @@ const stdSerializedResults: StdSerializedResults = {
     123: {},
   },
   req: {
+    params: {},
+    query: {},
     id: canBeUndefined('id'),
     method: 'GET',
     url: 'http://0.0.0.0',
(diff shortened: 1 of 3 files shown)


index.test-d.ts
```diff
@@ -138,6 +138,8 @@ const stdSerializedResults: StdSerializedResults = {
     123: {},
   },
   req: {
+    params: {},
```
jmealo, line 141: This fixes the typescript errors.

mcollina (approved):
lgtm

jmealo:
@mcollina: Do I need to do anything else to merge this? (Version bump, etc?)

Issue #299, opened 2023-10-08
Updating for v10


Ranieri93:
Hello everyone!
I would like to work on this, as needed in [issue #1762](https://github.com/pinojs/pino/issues/1762) and referenced [here](https://github.com/pinojs/koa-pino-logger/issues/63).

I guess here we don't need much explanation about the need for the update.
I should also mention that there is a dependency with [pino-std-serializers](https://github.com/pinojs/pino-std-serializers), so I guess I'll work on that too.

@jsumners

jsumners:
Thanks. I have created `next` branches here, in the koa module repo, and the standard serializers repo. Luckily, the standard serializers module is simple; it will probably only need some CI changes to eliminate old Node.js versions and some dependency bumps.

mcollina:
v10 shipped

PR #300, opened 2023-10-11
Updating for v9: update dependencies and CI

Ranieri93:
Hello everyone!

This PR is referencing this [issue](https://github.com/pinojs/pino-http/issues/299).

Quick question, should I leave `"pino-std-serializers": "^6.0.0"` like this? Or do we expect a major release after the update for v9?

Thank you in advance!

@jsumners

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
       contents: read
     strategy:
       matrix:
-        node-version: [14, 16, 18]
+        node-version: [18, 20]
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
--- a/package.json
+++ b/package.json
@@ -7,20 +7,20 @@
   "dependencies": {
     "get-caller-file": "^2.0.5",
     "pino": "^8.0.0",
-    "pino-std-serializers": "^6.0.0",
+    "pino-std-serializers": "git+https://github.com/pinojs/pino-std-serializers.git#next",
     "process-warning": "^2.0.0"
   },
   "devDependencies": {
+    "@fastify/pre-commit": "^2.0.2",
     "@types/node": "^20.1.0",
     "autocannon": "^7.3.0",
     "coveralls": "^3.0.0",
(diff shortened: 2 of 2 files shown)


Ranieri93:
> > Quick question, should I leave `"pino-std-serializers": "^6.0.0"` like this? Or do we expect a major release after the update for v9?
> 
> We will not be able to release new versions of ecosystem modules until `pino@9` is ready to be released. The updated ecosystem modules will be remain on their respective `next` branches until that time (see https://github.com/orgs/pinojs/projects/1). Any modules that have dependencies on each other should use git URLs to specify the dependencies until we are ready to publish actual versions.

I never used such a syntax, I researched a little bit and I hope this is the correct one, but correct me if I'm wrong!

Also, this is probably a stupid question but:
I'm seeing tests failing on node 20 because the coverage threshold is not met, what should be done in this case?
Because test response comes always with different uncovered line numbers, I'm a bit confused!

Thank you in advance as always for your response!

jsumners:
The primary issue is that this test is failing under Node 20: https://github.com/pinojs/pino-http/blob/9bf08025852013b5927b5b275ec5bc23fb38a015/test/test.js#L477-L517 

It's not clear to me why. I reduced the issue down to the following script that works correctly under Node 18 but returns a 400 response under node 20:

```js
'use strict'

const http = require('http')
const net = require('net')

const server = http.createServer(function (req, res) {
  res.statusCode = 200
  res.end('ok')
})

server.listen(0, '127.0.0.1', function (err) {
  if (err) {
    console.error(err)
    server.close()
    return
  }

  const client = net.connect(server.address().port, '127.0.0.1', () => {
    client.write('GET /delayed HTTP/1.1\r\n\r\n')
  })

  client.on('data', (data) => {
    console.log('data: ', data.toString())
    client.destroy()
    server.close()
  })
})
```

@mcollina any ideas here?

jsumners:
@Ranieri93 you should be able to rebase now to get passing tests.

jsumners (approved):
Looks good to me.

PR #304, opened 2023-10-23
fix: Pino options have a generic type

UndefinedBehaviour:
**Issue**
```
import { pinoHttp } from 'pino-http';

const httpLogger = pinoHttp({
    customLevels: {
        log: 20,
    },
});

const logger = httpLogger.logger;

function f(level: 'warn' | 'info' | 'log', message: string): void {
  logger[level](message);
}

```
This code is correct but currently throws a `ts(7053) : Element implicitly has an 'any' type` because TS can't see the custom level added in the configuration.


kibertoad (changes requested):
Please add tsd test for it!

UndefinedBehaviour:
Test added!

mcollina (approved):
lgtm

PR #318, opened 2023-12-19
fix: TypeScript for CustomLevels on parent pino instance

### LINKED ISSUE #317: TypeScript error when using customLevels on pino parent instance
Using pino@8.17.1 and pino-http@8.6.0 :

```
import pino from "pino";
import pinoHttp from "pino-http";

const log = pino({
  customLevels: {
    metric: 25
  }
});

export const logRequestsMiddleware = pinoHttp({
  logger: log
});
```

Build error: 
```
create-logger.ts:11:3 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(opts?: Options<IncomingMessage, ServerResponse<IncomingMessage>> | undefined, stream?: DestinationStream | undefined): HttpLogger<...>', gave the following error.
    Type 'Logger<"metric">' is not assignable to type 'Logger<never> | undefined'.
      Type 'Logger<"metric">' is not assignable to type 'Logger<never>'.
        Type 'Logger<"metric">' is not assignable to type 'LoggerExtras<never>'.
          Types of property 'onChild' are incompatible.
            Type 'OnChildCallback<"metric">' is not assignable to type 'OnChildCallback<never>'.
              Types of parameters 'child' and 'child' are incompatible.
                Type 'Logger<never>' is not assignable to type 'Logger<"metric">'.
                  Type 'Logger<never>' is not assignable to type 'LoggerExtras<"metric">'.
                    Types of property 'customLevels' are incompatible.
                      Property 'metric' is missing in type '{}' but required in type '{ metric: number; }'.
  Overload 2 of 2, '(stream?: DestinationStream | undefined): HttpLogger<IncomingMessage, ServerResponse<IncomingMessage>, never>', gave the following error.
    Object literal may only specify known properties, and 'logger' does not exist in type 'DestinationStream'.

11   logger: log
```

GeoffreyEmerson:
CustomLevels from pino parent instance should be passed through TypeScript variables.

Fixes #317

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,7 +12,7 @@ import { IncomingMessage, ServerResponse } from 'http';
 import pino from 'pino';
 import { err, req, res, SerializedError, SerializedRequest, SerializedResponse } from 'pino-std-serializers';
 
-declare function PinoHttp<IM = IncomingMessage, SR = ServerResponse, CustomLevels extends string = never>(opts?: Options<IM, SR>, stream?: pino.DestinationStream): HttpLogger<IM, SR, CustomLevels>;
+declare function PinoHttp<IM = IncomingMessage, SR = ServerResponse, CustomLevels extends string = never>(opts?: Options<IM, SR, CustomLevels>, stream?: pino.DestinationStream): HttpLogger<IM, SR, CustomLevels>;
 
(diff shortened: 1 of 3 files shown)


naseemkullah:
alternative approach, ptal: https://github.com/pinojs/pino-http/pull/319

mcollina (approved):
lgtm

PR #339, opened 2024-06-28
feat: add quietResLogger option

### LINKED ISSUE #235: [Proposal] support for quietResLogger (similar to quietReqLogger) - log full bindings at start of request instead at the end.

wms:
Applies similar logic as `quietReqLogger` to determine which logger instance to use.

Fixes #235


mcollina (approved):
lgtm

jsumners (approved):

PR #342, opened 2024-07-02
fix: allow using non-lowercased custom levels

chernodub:
The code failed to use `useLevel` when the custom level was, for example, camel-cased. 

I updated the test to reflect this occasional scenario and removed the unnecessary transformation of the log level.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -42,7 +42,7 @@ function pinoLogger (opts, stream) {
 
   function getValidLogLevel (level, defaultValue = 'info') {
     if (level && typeof level === 'string') {
-      const logLevel = level.trim().toLowerCase()
+      const logLevel = level.trim()
       if (validLogLevels.includes(logLevel) === true) {
         return logLevel
       }
--- a/test/test.js
+++ b/test/test.js
@@ -223,10 +223,10 @@ test('uses the log level passed in as an option, where the level is a custom one
   const logger = pinoHttp(
     {
       customLevels: {
-        custom: 25
+        infoCustom: 25
       },
-      useLevel: 'custom',
-      level: 'custom'
+      useLevel: 'infoCustom',
+      level: 'infoCustom'
     },
     dest
   )


test/test.js
```diff
@@ -223,10 +223,10 @@ test('uses the log level passed in as an option, where the level is a custom one
   const logger = pinoHttp(
     {
       customLevels: {
-        custom: 25
+        infoCustom: 25
       },
-      useLevel: 'custom',
-      level: 'custom'
+      useLevel: 'infoCustom',
+      level: 'infoCustom'
```
chernodub, line 229: If needed, I could duplicate the test for the lowercase scenario too. I'm assuming here that both cases are covered by this test now.

logger.js
```diff
@@ -42,7 +42,7 @@ function pinoLogger (opts, stream) {
 
   function getValidLogLevel (level, defaultValue = 'info') {
     if (level && typeof level === 'string') {
-      const logLevel = level.trim().toLowerCase()
```
jsumners, line 45: The purpose here is to compare names in a consistent manner.

jsumners (dismissed):
I don't understand what issue this is solving. With the current code, `fooBar` is equivalent to `foobar`.

logger.js
chernodub: Hi @jsumners, thanks for the response! 

**My reasoning for the change:** 

1. `validLogLevels` contains values that are not mapped to lowercased values, so the check will always fail for non-lowercased values. (like `infoCustom` example above). 
2. We could potentially prepare the `validLogLevels` by mapping each value to lowercase, so the check will result to `true`. After that, the `getValidLogLevel` will return a lowercased value, and the code of pino-http will try calling something like `logger.infocustom` which will also throw (because the valid call would be `logger.infoCustom`)

Pls let me know what you think, thanks

mcollina (approved):
lgtm

mcollina:
I think this is ok!

jsumners (approved):

PR #346, opened 2024-08-07
fix: allow CustomLevels in useLevel and customLogLevel

JulianCissen:
Fixed an issue where useLevel and customLogLevel Options parameters wouldn't accept CustomLevels.


mcollina (approved):
lgtm

PR #358, opened 2024-11-22
feat(types): add responseTime parameter to customErrorMessage

MarcusHong:
## Description
Add responseTime parameter to customErrorMessage type definition to allow access to the request's response time when creating custom error messages.

## Changes
- Updated TypeScript type definition for customErrorMessage to include responseTime parameter
- Response time is provided in milliseconds, consistent with pino-http's existing timing

## Example Usage
```typescript
const logger = pinoHttp({
  customErrorMessage: (req, res, err, responseTime) => {
    return `${err.message} (took ${responseTime}ms)`;
  }
});


mcollina:
Thanks for opening a PR! Can you please add a unit test? We use `tsd` for the types

Issue #361, opened 2024-12-09
Issue while testing - `res.allLogs` is not defined


todor-a:
I've encountered an issue while using jest-mock-extended - the issue is that when you create a response like this:
```
const response = mock<Response>();
```

This instantiates `allLogs` as a mocked function and [this check](https://github.com/pinojs/pino-http/blob/24af8277d9d255ba61494444675757eb79dcaecf/logger.js#L157) passes.

Does it make sense to change this line to `!Array.isArray(allLogs)`? Would be happy to submit a PR.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

todor-a:
Will do.

PR #362, opened 2024-12-09
feat: make check for allLogs more explicit

todor-a:
Closes https://github.com/pinojs/pino-http/issues/361.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -154,7 +154,7 @@ function pinoLogger (opts, stream) {
     if (!res.log) {
       res.log = responseLogger
     }
-    if (!res.allLogs) {
+    if (!Array.isArray(res.allLogs)) {
       res.allLogs = []
     }
     res.allLogs.push(responseLogger)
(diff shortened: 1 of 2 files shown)


logger.js
```diff
@@ -154,7 +154,7 @@ function pinoLogger (opts, stream) {
     if (!res.log) {
       res.log = responseLogger
     }
-    if (!res.allLogs) {
+    if (!Array.isArray(res.allLogs)) {
```
jsumners, line 157: The prior check was a coercive check because it was against a truthy/falsy value. I'd prefer it if we make the new check explicit:

```suggestion
    if (Array.isArray(res.allLogs) === false) {
```

jsumners (commented):
Looks good. Just a nit.

logger.js
todor-a: makes sense to me, fixed

jsumners (approved):

mcollina (approved):
lgtm

PR #372, opened 2025-07-29
fix(types): add responseTime to customErrorMessage callback

### LINKED ISSUE #359: Bug: `customErrorMessage` actually got passed 4 arguments, but it's ts declaration only has 3 arguments
`customErrorMessage` actually got passed 4 arguments:
 https://github.com/pinojs/pino-http/blob/8baaeec6fa2b53a7fe1c6bb948b3f46b7dbb6bdf/logger.js#L123
, but its ts declaration only has 3 arguments:
https://github.com/pinojs/pino-http/blob/8baaeec6fa2b53a7fe1c6bb948b3f46b7dbb6bdf/index.d.ts#L34

vikasrathod4299:
Fixes #359

Updated the `customErrorMessage` type definition to include the 4th argument `responseTime`.

This allows users to make full use of all parameters passed to `customErrorMessage`.

- Updated `Options` interface in `index.d.ts`
- Added test case to `index.test-d.ts` using 4 parameters


mcollina (approved):
lgtm

PR #426, opened 2026-08-01
fix: apply autoLogging.ignore to the request received log

Jaybhade:
### The problem

`autoLogging.ignore` is only consulted inside `onResFinished`. That silences the
"request completed" / "request errored" log, but the **"request received"** log — emitted
when `customReceivedMessage` or `customReceivedObject` is set — is written before any of
that, with no ignore check at all.

So the classic use case the option exists for, dropping health-check noise by url or
user-agent, still emits one line per ignored request as soon as received logging is on:

```js
const logger = pinoHttp({
  customReceivedMessage: (req) => `request received: ${req.url}`,
  autoLogging: { ignore: (req) => req.url === '/health' }
})
```

```
GET /health   →  {"msg":"request received: /health","req":{"url":"/health",...}}   ← ignored, still logged
GET /real     →  {"msg":"request received: /real",...}
                 {"msg":"request completed",...}
```

The completion log for `/health` is correctly suppressed; the received log is not.

### The fix

`ignore` is now consulted for the received log as well, via a shared
`isAutoLoggingIgnored()` helper so both call sites can't drift again.

The check is placed **after** the `receivedMessage !== undefined || onRequestReceivedObject !== undefined`
test, so it short-circuits: when received logging isn't configured — the default — `ignore`
is called exactly as often as it is today, and there is no added cost on the hot path.

### Two consequences worth a maintainer's eye

1. **`ignore` is called twice per request** when received logging is enabled — once per
   automatic log. That reads correctly to me ("is this log ignored?"), and it's the only way
   to gate a log that happens at a different moment, but it does mean an `ignore` with side
   effects now runs twice.
2. **An `ignore` that filters on `res.statusCode` sees the default value at received time**,
   because the response hasn't been written yet. Such a predicate will suppress the received
   log based on a placeholder status, while the completion log is still evaluated correctly
   against the real one. This is inherent to gating a log that is emitted before the response
   exists rather than something this patch chooses; I've documented it on the option instead
   of trying to paper over it. Happy to change the approach if you'd rather the received log
   stay ungated for status-code predicates.

Both are called out in the README change.

### Verification

Three tests added next to the existing `autoLogging.ignore` tests:

- `autoLogging.ignore also skips the request received log` — `customReceivedMessage`, ignored url, expects no output. **Fails on master**, passes with the fix.
- `autoLogging.ignore also skips the request received object` — same for `customReceivedObject`. **Fails on master**, passes with the fix.
- `autoLogging.ignore does not skip the request received log of a request it does not ignore` — guards against over-suppression by asserting a non-ignored request still logs both received *and* completed. Passes on master and with the fix, so it pins the behaviour that must not regress.

`npm test` (eslint + `node --test` + c8 + `tsc`/`tsd`): **67 passing, 0 failing, coverage still 100%** on `logger.js`.

No behaviour change for anyone not using `customReceivedMessage` / `customReceivedObject`.


mcollina (approved):
lgtm

